PR #313, opened 2012-08-21
Added AttemptPostFilesWithoutBuffering flag for large file uploads

BContario:
Exposed a new flag in the RestRequest interface called
AttemptPostFilesWithoutBuffering.  This allows the client to try to
upload files WITHOUT buffering the entire file in memory.  This is
important when trying to upload because extremely large files or
simultaneous uploads could easily exhaust system memory.


mdissel:
Is the memory consumption still a problem with large files (the stream is loaded in memory)?

chenzhekl:
Don't understand why this PR was closed. Sending large files using POST is still a pain.

Issue #401, opened 2013-08-21
Allow setting CachingPolicy for the request


galmeida:
That's it. It would be nice to have a way to set a CachingPolicy for request.

dgreenbean:
I just created a pull request for a more limited version of the original change proposed. This does not change the default functionality, but it does add the ability to set the cache policy on a client.

This is important, as the only workaround currently is to set the cache policy globally for a process. In general, it is better to limit the scope of the policy to a specific client wrapper, i.e. only specific requests.

elmerillion:
This was merged but doesn't appear to be active due to the "#if FRAMEWORK" checks. Is there a compiler directive missing?

dgreenbean:
@elmerillion I see the change when I pull version 105.2.3 from NuGet.  This change will, of course, only work for builds with the FRAMEWORK variable defined in the project file, e.g.: https://github.com/restsharp/RestSharp/blob/master/RestSharp/RestSharp.Net35.csproj#L44

This means that if you need it for Silverlight, Windows Phone, iOS, etc., support would need to be added in a separate pull request.  I was very conservative with my pull request to limit the impact of the change.

elmerillion:
@dgreenbean thanks, I think I had a pre-merge version loaded for some reason and couldn't find the property. My mistake!

alexeyzimarev:
I assume since the PR was accepted, the issue is now resolved.

PR #403, opened 2013-08-21
fix for issue #401 allow setting CachingPolicy for the request

ghost:
This is my first pull request in github - please correct me if anything is wrong.
I don't know how to attach this merge request with issue: https://github.com/restsharp/RestSharp/issues/401


duhast:
Hello guys!
Since this is not completed, can anybody tell me how I can disable caching for RestSharp/HttpWebRequest from my code (without modifying RestSharp sources)?

dgreenbean:
I just created a pull request for a more limited version of the original change proposed.  This does not change the default functionality, but it does add the ability to set the cache policy on a client.

This is important, as the only workaround currently is to set the cache policy globally for a process.  In general, it is better to limit the scope of the policy to a specific client wrapper, i.e. only specific requests.

PR #464, opened 2013-11-13
Host header for requests can be specified explicitly

### LINKED ISSUE #462: Add ability to customize Host header
By default the Host header is based on URL provided for the request. In some circumstances it's needed to specify host explicitly. For example, in case of port forwarding we need to specify exact IP address in URL and the Host header should be specified explicitly.

This behavior is supported by **Host** property of **HttpWebRequest** class from .NET 4. But it's not supported by RestSharp yet.

alxbog:
Uses HttpWebRequest.Host property (.NET 4). Fixes #462.


cuibty:
+1

PR #469, opened 2013-11-15
Streaming Support

petertoviskes:
Add Streaming support through FileParameter.


hallem:
@petertoviskes can you add unit tests?

PR #592, opened 2014-10-31
Make the attributes search work the same as for elements

lycilph:


lycilph:
I don't know how to update this request so i will close this and make a new one

PR #618, opened 2014-12-19
Added and updated the Xamarin projects

mattleibow:
- Added Xamarin.iOS (Unified API)
 - Updated MonoTouch (Classic API)
 - Changed MonoDroid to Android as this is more correct and supports the newer build/projects


hallem:
Matt can you send me a detailed description of what the removals were for either her or in a separate email? I just want to make sure I completely understand what you did before I merge everything in.  This isn't an area of expertise for me so I need to be careful.

mattleibow:
no Prob,

The removals were two parts:

 - the old solution files: RestSharp.MonoDroid.sln and RestSharp.MonoTouch.sln are now in the RestSharp.Xamarin.sln solution.
 - the duplicate code "ResponseStatusExtensions.cs" are just copies of the same file in the RestSharp/Extensions directory. I just deleted the copies and use the one in the main directory.

The major changes in the MonoTouch project file was just because the file was quite out of date.

Hope this helps.

Issue #552, opened 2014-07-22
NullReferenceException when calling ExecuteTaskAsync


Gimly:
I'm fighting with the `TaskAsync` methods of RestSharp, and losing for now.

First, there seems to be a big inconsistency between the "standard" version of the Execute method and the asynchronous one.

I've tried this code:
```C#
// Connect to any non-existing IP address
var client = new RestClient("http://192.168.1.200:8001");
var request = new RestRequest("/")
{
   RequestFormat = DataFormat.Json,
   Method = Method.GET
};
var response = client.Execute(request);
```
The code works, I get no exception and the response has simply a `ResponseStatus == Error` and the `ErrorMessage` corresponding to the error (could not connect). So far so good.

Now, let's try the same thing with the Async methods:
```C#
// Connect to any non-existing IP address
var client = new RestClient("http://192.168.1.200:8001");
var request = new RestRequest("/")
{
   RequestFormat = DataFormat.Json,
   Method = Method.GET
};

var response = await client.ExecuteTaskAsync(request);
```
I would expect the same behavior as the "standard" call, no exception and a `ResponseStatus` in error. Except, I get a `WebException` that is thrown because it couldn't connect to the server.
OK, it's inconsistent, but I can still work around that, I'll catch the exception and manage it.

Now, let's try to use the generic verison, `ExecuteTaskAsync<T>` which parses the result into a class.

```C#
// Connect to any non-existing IP address
var client = new RestClient("http://192.168.1.200:8001");
var request = new RestRequest("/")
{
   RequestFormat = DataFormat.Json,
   Method = Method.GET
};

var response = await client.ExecuteTaskAsync<StupidClass>(request);
```

In that case, everything crashes and I get a `NullReferenceException` coming from inside RestSharp's code, seemingly from the `DeserializeResponse` method. For some reason (the code being on another thread I guess) I cannot even catch it and it crashes the whole application.

Stack trace:
```
   at RestSharp.RestClient.<>c__DisplayClass15`1.<ExecuteTaskAsync>b__12(IRestResponse`1 response, RestRequestAsyncHandle _)
   at RestSharp.RestClient.DeserializeResponse[T](IRestRequest request, Action`2 callback, IRestResponse response, RestRequestAsyncHandle asyncHandle)
   at RestSharp.RestClient.<>c__DisplayClassa`1.<ExecuteAsync>b__9(IRestResponse response, RestRequestAsyncHandle asyncHandle)
[...]

Issue #466, opened 2013-11-13
ExecuteTaskAsync null response is causing an unhandled exception


ghost:
I'm testing using my app offline by switching off my server. If I call ExecuteTaskAsync<T> I'm getting an unhandled exception since the server is unavailable and the response is null. 

The error arises in RestClient.Async.cs,

var async = ExecuteAsync<T>(request, (response, _) =>
{
....
});

Would it make sense to handle the case where response is null e.g. 

else if (response == null)
{
    taskCompletionSource.TrySetException(new NullReferenceException());
}

or is there a better way this should be handled ?

alexeyzimarev:
This seems to be fixed by that PR.

PR #628, opened 2015-01-08
Fix to issue in #466 and others. ExecuteTaskAsync should not throw exceptions

jacdevos:
This issue seems to plague a lot of people.

Execute (sync) and ExecuteAsync handle exceptions by swallowing them and setting the Error properties on the response object. This change makes ExecuteTaskAsync also swallow its errors by NOT invoking  taskCompletionSource.TrySetException(response.ErrorException) in RestClient.Async.cs.

The change in Http.Async.cs was simply to populate the ErrorException with a ResponseStatus.ToWebException() if empty, making it even more consistent with Execute (sync).

Most of the code changes were changes to tests that expected ExecuteTaskAsync to throw exceptions.

### DIFF
--- a/RestSharp.IntegrationTests/AsyncTests.cs
+++ b/RestSharp.IntegrationTests/AsyncTests.cs
@@ -87,30 +87,16 @@ namespace RestSharp.IntegrationTests
                 var client = new RestClient(baseUrl);
                 var request = new RestRequest("success");
 
-                request.OnBeforeDeserialization += response =>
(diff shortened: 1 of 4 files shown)


jacdevos:
Hi Michael,

Let me know when you get a chance to look at pulling this in. Just a correction, it fixes #552 and related issues rather than #466.

Hope it helps.

hallem:
Thanks for the PR.  I'll try and take a look at it this weekend.

PR #632, opened 2015-01-20
Http.Sync HttpWebRequest.Abort() after execution > Prevents requests hang in Android

raste:
Hello, first of all, sorry if I haven't done something correctly (first pull request ever).

On Android (version doesn't matter) the connections start to hang (never return result -> which leads to time out) after 3rd sequential failure on RestClient(.Sync).Execute().

Example test case: Send requests to HTTPS url's whit set invalid date on the client (it must be outside of the valid boundaries on the SSL certificate on the server) (so that the handshake with the server certificate fails)

Concrete example: set date on the client (android device/emulator) January 2014
Make HTTPS requests with the following code:

```
HttpWebRequest request = (HttpWebRequest)HttpWebRequest.Create(new Uri("https://app.operadds.com/api/GetServerUTCTimestamp"));

request.ContentType = "application/json";
request.Method = "GET";

try{
    WebResponse response = request.GetResponse();
}
catch{}
```

On the first 3 tries exception will be thrown and you will land in the catch block, with SSL certificate handshake error. (the exception reason doesn't matter)
On the 4th and every other try `request.GetResponse()` will not return result (not even exception will be thrown). When this happens, the only option is to restart the application... (even if you fix the date, the problem will persist)

Even though the test case doesn't use RestSharp, the library does the same calls and the same result is presented.

The fix, which seems to work is to just call `request.Abort();` after exception

*Note: I could not reproduce the bug on Windows Desktop and iOS, haven't tested on Windows Phone*


hallem:
Since this is an android specific fix, and it hasn't been tested anywhere else, we should wrap it in an if def statement that specifically targets android.  Make sense?

raste:
You are right, is this the correct if/def statement?

PR #637, opened 2015-01-28
Fixes #636 - Header of 'Range' now supports a rangeSpecifier

### LINKED ISSUE #636: AddHeader with rangeSpecifier
RestSharp currently only allows the Range header with the "bytes" specifier:
Range: bytes=1-2000

However, the HTTP spec allows for any specifier to be used with Range:
Range: records=1-200
http://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.35.2

An update to AddRange() in Http.cs has been suggested here:
https://groups.google.com/forum/?hl=en#!topic/restsharp/VKTH6sg9b_4

drusellers:
Here's a quick attempt to fix the #636 AddHeader

- [ ] I am not sure how to best test this (I monkey tested it, but that doesn't count)

### DIFF
--- a/RestSharp.Tests/AddRangeTests.cs
+++ b/RestSharp.Tests/AddRangeTests.cs
@@ -1,0 +1,17 @@
+﻿namespace RestSharp.Tests
+{
+    using Xunit;
+
+    public class AddRangeTests
+    {
+        [Fact]
+        public void ShouldParseOutRangeSpecifier()
+        {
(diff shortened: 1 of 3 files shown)


hallem:
Thanks for the PR.  What issue are you trying to solve with this?  Its not clear from your test or your description.

drusellers:
Trying to close issue #636

PR #639, opened 2015-02-11
added IRestRequest AddFile(string name, string path, string contentType)

petrce:
now it's possible to upload a file with a correct content-type


hallem:
Thanks for the PR.  Can you add a test for this?

petrce:
Not a problem his give me a week, I am on vacation.

hallem:
I merged it in with some changes.  Check them out when you get back and make sure it's still working for your use case.

PR #642, opened 2015-02-14
DotNetXml Serialization Settings

EddieDL:
I have modified both DotNetXmlDeserializer and DotNetXmlSerializer to
allow for a full configuration of the Xml serialization.  I kept the
Encoding property but I am not sure I should have.

This is my first time doing anything on Git so please be kind :)


hallem:
Eddie what was the issue you were trying to fix with this?  Can you provide a breaking test and how this fixes the issue?

EddieDL:
My problem was not so much a bug as much as a feature gap.  In my case the REST service I am trying to connect to has a few special requirements for it's XML.  For example it can't have an XML declaration in it's request and the way the Namespace property was being handled it forced all of my XML to have a qualifier, which this service did not support.  

After working on this some more, I am not convinced that my solution is the best one but the main goal of my change was to allow the caller to have more control over the XML serialization and deserialization.  If anyone thinks this might be useful, I will further refine my change to include some of my other changes I made in my private project.

If you would like an example test I can try to create something.  Just let me know if it would help.

hallem:
Eddie I'm not sure that we should merge this in then.  It sounds like this might be something very specific to the service that you're interfacing with.  I'd hate to put something in that technically doesn't follow the XML standards.  What do you think?

hallem:
closed as not appropriate for the library due to too specific of a use case

PR #644, opened 2015-02-18
Adding support for deserializing ISO 8601 durations as TimeSpan in JSON

briangweber:
It appears that ServiceStack's default serialization for TimeSpan is ISO 8601, which causes an exception during deserialization in RestSharp. XmlConvert handles this format, so we'll attempt normal TimeSpan.Parse and failing that, attempt XmlConvert.


hallem:
This looks good so far.  Thanks for the tests and the PR.  I'll review it this week and get it merged in.

PR #646, opened 2015-02-26
Make the attributes search work the same as for elements

lycilph:
I closed the previous request and i am adding a new one with unit test, that show the issue


hallem:
Thanks for doing that.  I'll review everything and get it merged in soon.

PR #658, opened 2015-03-23
Attempt to fix AppVeyor integration test issues with XUnit.

issafram:
Using `Environment.NewLine` instead of hardcoding `\r\n`.  Hoping this fixes any previous issues with AppVeyor.  Only way to test for me is to push the PR.

Also removed all references to NUnit.


issafram:
AppVeryor build seems to have failed due to a separate issue in `SimpleServer`.  I'll look into it but the unit tests passed with no issue.

PR #660, opened 2015-03-25
SimpleServer Cleanup With Added Try Catches

issafram:
I hope this will prevent some of the random failing of unit tests we see on AppVeyor.

Threads can always be tricky but this could be a great first step.


issafram:
@hallem, can you take a look at this?

PR #665, opened 2015-03-27
Allow additional file Content-Disposition data

jholland918:
Allow passing additional Content-Disposition data for files. I'm trying to use RestSharp for my DocuSign integration project. Some of DocuSign's REST API endpoints require custom values to be added to the Content-Disposition when uploading files.


jholland918:
Eh, not sure why this failed the AppVeyor build. I'm also not sure if the HTTP gods allow custom key/value pairs to be added to the Content-Disposition in the standard anyway. I'll just close this and just stuff what I need in the file or filename parameters for now.

PR #669, opened 2015-04-07
Merged Portable project to Master Branch

IlSocio:
I moved the portable project to the master branch, by including the Portable project in the main solution and adding the PORTABLE symbol.
The main aim is to keep the Portable version updated by using a single codebase where possible.

Indeed, I noticed that most of the Portable files are identical to the Non-Portable ones, but the copies on the portable branch are currently outdated (eg: Authenticators / Serializers / Deserializers / SimpleJson / Extensions / Validation) so, it makes sense to me to change those files to linked files.

Imho, the Portable branch should be removed.


IlSocio:
I can't understand why the Portable build is failing :(
I included the "microsoft http client library" nuget package, but it seems that AppVeyor can't find the http namespace...

hallem:
Thanks for doing this.  The build failed though, take a look at the build report and update where necessary please so it can be merged in.

PR #673, opened 2015-04-09
DynamicObject Support for .Net4+

icyflash:
Now you can use json.name instead of json["name"]


issafram:
PR #660 might help with this build failure.  We seem to have a threading issue in our tests.

PR #674, opened 2015-04-09
Merged Portable project to Master Branch 

IlSocio:
Trying again to create a pull request after some cleanup...

More important, I migrated all the projects from msbuild package restore to nuget package restore (I suspect msbuild package restore was the cause of the previous failure during build process)
https://docs.nuget.org/consume/package-restore/migrating-to-automatic-package-restore
https://docs.nuget.org/consume/package-restore
http://blog.davidebbo.com/2014/01/the-right-way-to-restore-nuget-packages.html

I moved the portable project to the master branch, by including the Portable project in the main solution and adding the PORTABLE symbol.
The main aim is to keep the Portable version updated by using a single codebase where possible.

Indeed, I noticed that most of the Portable files are identical to the Non-Portable ones, but the copies on the portable branch are currently outdated (eg: Authenticators / Serializers / Deserializers / SimpleJson / Extensions / Validation) so, it makes sense to me to change those files to linked files.

Imho, the Portable branch should be removed.


hallem:
@IlSocio are you still working on this?

IlSocio:
Yes, but there are a lot of tests which are currrently failing and I'll need some spare time to take a deep look to it...

PR #676, opened 2015-04-09
Migrated from "MSBuild Package Restore" to "NuGet Automatic Package Rest...

IlSocio:
https://docs.nuget.org/consume/package-restore/migrating-to-automatic-package-restore


hallem:
decided not to do this

PR #677, opened 2015-04-11
Namespace Cleanup

issafram:
Cleanup the namespaces for the codebase.

Main changes was changing namespace `RestSharp` to `RestSharp.Authenticators` for all classes in the Authenticators folder.

Also renamed namespace `RestSharp.Contrib` to `RestSharp.Extensions.MonoHttp`.

### DIFF
--- a/RestSharp.Tests/NamespacedXmlTests.cs
+++ b/RestSharp.Tests/NamespacedXmlTests.cs
@@ -18,11 +18,14 @@ using System;
 using System.Collections.Generic;
 using System.Xml.Linq;
 using RestSharp.Deserializers;
-using RestSharp.Tests.SampleClasses.Lastfm;
+
 using Xunit;
 
 namespace RestSharp.Tests
 {
+    using RestSharp.Tests.SampleClasses;
+    using RestSharp.Tests.SampleClasses.LastFm;
+
     public class NamespacedXmlTests
     {
         private const string GuidString = "AC1FC4BC-087A-4242-B8EE-C53EBE9887A5";
--- a/RestSharp.Tests/SampleClasses/LastFm/Lastfm.cs
+++ b/RestSharp.Tests/SampleClasses/LastFm/Lastfm.cs
@@ -1,8 +1,8 @@
-﻿using System;
-using System.Collections.Generic;
-
-namespace RestSharp.Tests.SampleClasses.Lastfm
+﻿namespace RestSharp.Tests.SampleClasses.LastFm
 {
+    using System;
+    using System.Collections.Generic;
+
     public class Event : LastfmBase
     {
         public string id { get; set; }
--- a/RestSharp.Tests/SampleClasses/misc.cs
+++ b/RestSharp.Tests/SampleClasses/misc.cs
@@ -14,12 +14,13 @@
 //   limitations under the License. 
 #endregion
 
-using System;
-using System.Collections.Generic;
-using RestSharp.Serializers;
-
-namespace RestSharp.Tests
+namespace RestSharp.Tests.SampleClasses
 {
+    using System;
+    using System.Collections.Generic;
+
+    using RestSharp.Serializers;
+
     public class PersonForXml
     {
         public string Name { get; set; }
@@ -40,11 +41,11 @@ namespace RestSharp.Tests
 
         protected string Ignore { get; set; }
 
-        public string IgnoreProxy { get { return Ignore; } }
+        public string IgnoreProxy { get { return this.Ignore; } }
 
         protected string ReadOnly { get { return null; } }
 
-        public string ReadOnlyProxy { get { return ReadOnly; } }
+        public string ReadOnlyProxy { get { return this.ReadOnly; } }
 
         public FoeList Foes { get; set; }
 
@@ -102,11 +103,11 @@ namespace RestSharp.Tests
 
         protected string Ignore { get; set; }
 
-        public string IgnoreProxy { get { return Ignore; } }
+        public string IgnoreProxy { get { return this.Ignore; } }
 
         protected string ReadOnly { get { return null; } }
 
-        public string ReadOnlyProxy { get { return ReadOnly; } }
+        public string ReadOnlyProxy { get { return this.ReadOnly; } }
 
         public Dictionary<string, Foe> Foes { get; set; }
 
--- a/RestSharp.Tests/SampleClasses/nullables.cs
+++ b/RestSharp.Tests/SampleClasses/nullables.cs
@@ -1,7 +1,7 @@
-﻿using System;
-
-namespace RestSharp.Tests
+﻿namespace RestSharp.Tests.SampleClasses
 {
+    using System;
+
     public class NullableValues
     {
         public int? Id { get; set; }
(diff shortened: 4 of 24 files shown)


hallem:
@issafram This is a great change but it would be a breaking one.  Other than clean up was there a reason for it?

issafram:
@hallem Well because it is proper.  To be fair,  it doesn't break functionality in any way.  Yes namespaces change, so if someone were updating their NuGet packages, existing code would have to change if they were using the old namespaces.

hallem:
@issafram I agree with you and it will get merged in.  I will probably hold it for one release because it would require potential code changes for users.  Just wanted to make sure that I wasn't missing something else.

RestSharp.Tests/NamespacedXmlTests.cs
```diff
@@ -18,11 +18,14 @@
 using System.Collections.Generic;
 using System.Xml.Linq;
 using RestSharp.Deserializers;
-using RestSharp.Tests.SampleClasses.Lastfm;
+
 using Xunit;
 
 namespace RestSharp.Tests
 {
+    using RestSharp.Tests.SampleClasses;
```
hallem: Let's move these using statements back outside the namespace

RestSharp.Tests/SampleClasses/LastFm/Lastfm.cs
```diff
@@ -1,8 +1,8 @@
-﻿using System;
-using System.Collections.Generic;
-
-namespace RestSharp.Tests.SampleClasses.Lastfm
+﻿namespace RestSharp.Tests.SampleClasses.LastFm
 {
+    using System;
```
hallem: Let's move these using statements back outside the namespace

RestSharp.Tests/SampleClasses/misc.cs
```diff
@@ -14,12 +14,13 @@
...
 #endregion
 
-using System;
-using System.Collections.Generic;
-using RestSharp.Serializers;
-
-namespace RestSharp.Tests
+namespace RestSharp.Tests.SampleClasses
 {
+    using System;
```
hallem: Let's move these using statements back outside the namespace

RestSharp.Tests/SampleClasses/nullables.cs
```diff
@@ -1,7 +1,7 @@
-﻿using System;
-
-namespace RestSharp.Tests
+﻿namespace RestSharp.Tests.SampleClasses
 {
+    using System;
```
hallem: Let's move these using statements back outside the namespace

RestSharp.Tests/SerializerTests.cs
```diff
@@ -23,6 +23,8 @@
 
 namespace RestSharp.Tests
 {
+    using RestSharp.Tests.SampleClasses;
```
hallem: Let's move these using statements back outside the namespace

RestSharp/Authenticators/HttpBasicAuthenticator.cs
```diff
@@ -14,12 +14,12 @@
...
 #endregion
 
-using System;
-using System.Linq;
-using System.Text;
-
-namespace RestSharp
+namespace RestSharp.Authenticators
 {
+    using System;
```
hallem: Let's move these using statements back outside the namespace

RestSharp/Authenticators/HttpBasicAuthenticator.cs
```diff
@@ -27,7 +27,7 @@ public class HttpBasicAuthenticator : IAuthenticator
         public HttpBasicAuthenticator(string username, string password)
         {
             var token = Convert.ToBase64String(Encoding.UTF8.GetBytes(string.Format("{0}:{1}", username, password)));
-            _authHeader = string.Format("Basic {0}", token);
+            this._authHeader = string.Format("Basic {0}", token);
```
hallem: if you're going to use "this." then remove the underscore please

RestSharp/Authenticators/NtlmAuthenticator.cs
```diff
@@ -14,13 +14,14 @@
...
 
-using System;
-using System.Net;
 
 #if FRAMEWORK
 
-namespace RestSharp
+namespace RestSharp.Authenticators
 {
+    using System;
```
hallem: Let's move these using statements back outside the namespace

RestSharp/Authenticators/OAuth2Authenticator.cs
```diff
@@ -14,11 +14,11 @@
 //   limitations under the License. 
 #endregion
 
-using System;
-using System.Linq;
-
-namespace RestSharp
+namespace RestSharp.Authenticators
 {
+    using System;
```
hallem: Let's move these using statements back outside the namespace

RestSharp/Extensions/MonoHttp/HttpUtility.cs
```diff
@@ -29,15 +29,14 @@
...
-using System.Collections;
-using System.Collections.Generic;
-using System.Collections.Specialized;
-using System.IO;
-using System.Text;
-
-namespace RestSharp.Contrib
+namespace RestSharp.Extensions.MonoHttp
 {
+    using System;
```
hallem: Let's move these using statements back outside the namespace

RestSharp/IRestClient.cs
```diff
@@ -28,6 +28,8 @@
 
 namespace RestSharp
 {
+    using RestSharp.Authenticators;
```
hallem: Let's move these using statements back outside the namespace

hallem:
just went through this, in general it looks good.  I have a few concerns:
1. The using statements should be outside the namespace declaration
2. If you're going to use "this." then make sure the member doesn't start with an underscore
3. For the using statements you removed, how did you determine they could be taken out? Some of the projects use things that aren't needed by others and since all the source files are shared we have to be very careful about removing using statements.  Especially for the mobile and portable stuff as they aren't apart of the overall build process.

issafram:
I can make those changes @hallem.

Number 1 specifically: I was trying to follow some StyleCop rules, such as http://www.stylecop.com/docs/SA1200.html.  Maybe we should have a set standard or a StyleCop rules file?  Just a thought.

Number 2:  That was to be a part of a bigger rename to remove the underscores.

Number 3:  I didn't keep that in mind.  Perhaps I might need to go through them and add some back.  Or find a way to determine what is valid/invalid.  I was using ReSharper to help do that.

hallem:
I agree with the StyleCop rules.  I'm a big fan so I'll look into getting them applied.  ReSharper is a great tool and I use it all the time, I only found out about the using directive thing by doing exactly what you did and removed some of the ones that weren't needed according to it only to find that they were later.  There is other clean up that should happen then though like the proper if def statements to make sure it's a known thing.

hallem:
going to do this at a later date closing this for now as it's too far out of sync

issafram:
Woah...

hallem:
sorry man, it's too far out of sync, i'm going to still migrate the namespaces as you have suggested just as a future item.  if you want to take a crack at it with the latest code, you're welcome to.

issafram:
I understand.

I'll do a new PR soon.

Thanks for still maintaining the project.
I appreciate it very much.

Best email notifications I've received all day.

hallem:
no problem, i've been away for a bit but trying to catch up now

hallem:
got this into the new release today.  check things out.  also more nuget targets

issafram:
Thank you for this.

PR #681, opened 2015-04-24
Deserializing an unknown enum value (numeric/alphanumeric) should return default value

kondaskondas:
`
public enum Disposition
    {
        Friendly,
        SoSo,
        SteerVeryClear
    }
`
When you deserialize an invalid numeric value eg: `999` it returns `Friendly` one would expect that deserializing and invalid alphanumeric value would have the same behavior but instead throws and exception (tries to convert "string" to  `int`).


seertenedos:
I would think this should be a setting.  Some cases you want Unknown like normal and others you want to know the missing value so you can fix it. Improving the error message with a try catch block that puts the value and enum name in the exception could also be a good solution.

```
public static object FindEnumValue(this Type type, string value, CultureInfo culture)
        {
            Enum @enum = Enumerable.FirstOrDefault<Enum>(Enumerable.Cast<Enum>((IEnumerable)Enum.GetValues(type)), (Func<Enum, bool>)(v => Enumerable.Contains<string>(StringExtensions.GetNameVariants(((object)v).ToString(), culture), value, (IEqualityComparer<string>)StringComparer.Create(culture, true))));
            if (@enum == null)
            {
                try
                {
                    object obj = Convert.ChangeType((object) value, Enum.GetUnderlyingType(type),
                        (IFormatProvider) culture);
                    if (obj != null && Enum.IsDefined(type, obj))
                        @enum = (Enum) Enum.ToObject(type, obj);
                }
                catch (Exception)
                {
                    throw new Exception(string.Format("Value '{0}' could not be matched to a value in enum '{1}'", value, type.Name));
                }

            }
            return (object)@enum;
        }
```

hallem:
Actually this behavior is consistent with how C# enums work.  An enums default value is 0 whether it has a member that equates to 0 or not.  The reason yours is defaulting to "Friendly" is because that holds the value of 0 since it's the first in the list and you haven't explicitly set the values in the enum.

PR #682, opened 2015-04-29
iOS project type update

louiecaulfield:
Makes RestSharp compile using Xamarin.iOS instead of monotouch


hallem:
there is a project specifically targeting ios already.

PR #685, opened 2015-05-10
Enhance Debug experience w/ RestResponse

drusellers:
One of the biggest pet peeves of RestSharp is the lack of Debugger Visualizations. This pull request is an attempt to show that a simple change can make a huge difference in the debug experience.

##Current
image

##With PR
image

The big change that I want is to easily see the HTTP Status Code, beyond that I'm open to any and all suggestions and feedback.

Thank you.

### DIFF
--- a/RestSharp/RestResponse.cs
+++ b/RestSharp/RestResponse.cs
@@ -18,12 +18,15 @@ using System;
 using System.Collections.Generic;
 using System.Net;
 using RestSharp.Extensions;
+using System.Diagnostics;
 
 namespace RestSharp
 {
+
     /// <summary>
     /// Base class for common properties shared by RestResponse and RestResponse[[T]]
     /// </summary>
+    [DebuggerDisplay("{DebuggerDisplay()}")]
     public abstract class RestResponseBase
     {
         private string _content;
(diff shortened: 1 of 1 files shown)


drusellers:
I looked over the details, I'm not sure why my build failed. Any hits so I can have a green build? Also, any feedback on the PR?

hallem:
Hey so the build probably failed for some bizarre reason.  I'll take a look and see what happened but I pull and test the PR's locally most times so I'm not worried.

As for the PR, it looks good to me.  The format may need to be slightly configurable to make it more useful to a wider audience but that can be done later.

Only question I'd have is does Silverlight, Windows Phone, and the other platforms support this functionality?  If not we'll need to add some #ifdef statements to limit to just the framework.

drusellers:
@hallem I haven't ever done work on those platforms so I'm not sure. :( But its a VS type feature so I wouldn't _expect_ that its an issue - but I have no real idea. As far as configurable - I have never seen that done, I usually just try and expose the most relevant details - but I could totally see that what I think is universal is not _really_ universal. :)

hallem:
decided that this wasn't a good idea to add.  to specific of a use case.

drusellers:
@hallem how is seeing the request code too specific of a use case? I can fully appreciate that there are a lot of data points that we all might want to see, but when I'm debugging a request / response its bothersome to have to dig into the request in the debugger to even see the HTTP Status Code. So, my request is can we at least expose the HTTP Status Code?

hallem:
fair enough let me look at it again

drusellers:
Thank You! That will save me a lot of time.

Issue #623, opened 2015-01-01
Package doesn't target MonoAndroid, Version=v5.0


rmtuckerphx:
This is the NuGet error I get when trying to install RestSharp 105.0.1:

Installing 'RestSharp 105.0.1'.
Successfully installed 'RestSharp 105.0.1'.
Adding 'RestSharp 105.0.1' to iotree.
Uninstalling 'RestSharp 105.0.1'.
Successfully uninstalled 'RestSharp 105.0.1'.
Install failed. Rolling back...
Could not install package 'RestSharp 105.0.1'. You are trying to install this package into a project that targets 'MonoAndroid,Version=v5.0', but the package does not contain any assembly references or content files that are compatible with that framework. For more information, contact the package author.

hallem:
The nuget package doesn't support the PCL or the Mono frameworks.  You need to clone the repo and compile it yourself for right now.

vasundharab:
You can install it form the xamarin component store

hallem:
just published a version that targets mono android (xamarin) to nuget

PR #710, opened 2015-07-02
Changes to support generating a nuget including MonoAndroid.

josea:
I changed the build.bat so if one choses you can generate the NuGet including the MonoAndroid support. 

It can support batch compiling if you have a Xamarin Business Level licensing. Otherwise, the user has to manually build the RestSharp.Android project (from VisualStudio) and run the batch. 

Usage: 
* `build.bat Release MonoAndroid`: it will build the nuget assuming the user manually built the RestSharp.Android project.

* `build.bat Release MonoAndroid-Business`: it will build it and include it in the result. 

I do not know your actual workflow to publish new versions of the nuget, but even if you ignore these options in your build process at least it would be easy for somebody to clone the repository and generate the nuget locally (and add it to a local/internal nuget repository).

This actually (partially) addresses this issue: https://github.com/restsharp/RestSharp/issues/623


hallem:
did this manually in the wip branch.  nuget targets coming very soon.

PR #734, opened 2015-08-19
Pseudo PCL support

KraigM:
I added Pseudo PCL support to RestSharp (aka AdvancedPCL or Bait/Switch PCL).  This will allow full PCL projects to build against RestSharp but the nuget package will swap in the real implementation for the actual application.  

I started with the stubs file created by @mattleibow in RestSharp.Portable and opted in files that were already compatible without and changes to the existing code.  Other files may be close (only needing to add to the #if's that are already there), but these were able to be added without effecting existing code.

Obtained Code From : https://github.com/mattleibow/RestSharp.Portable/blob/master/generated/stubs.generated.cs (same Apache 2.0 License)


hallem:
i'm not sure i understand how this works or why it's needed

KraigM:
It allows you to build a PCL only library that uses RestSharp. Even though the PCL version of RestSharp doesn't work, the working version will be swapped in when the app/program/etc pulls in the platform specific version of RestSharp.  In other words, one full PCL dll can be released that relies on RestSharp, despite RestSharp not being full PCL supported.

Many people have started using this trick recently to support some amazing things like platform agnostic cryptography with full native speeds.

Here are a articles on the topic:
http://log.paulbetts.org/the-bait-and-switch-pcl-trick/
http://ericsink.com/entries/pcl_bait_and_switch.html

batkuip:
This would be very handy indeed.  Atm I can't use RestSharp in our Xamarin project as all the logic is in a Core PCL library shared across platforms - this is a very common approach when targeting multiple platforms.

amis92:
Please please please accept this pull request and release on NuGet, it'd be absolutely awesome! As earlier mentioned, the technique is used in many projects and works.

PCL-compat RestSharp NuGet is a dream to many (especially) mobile devs.

hallem:
i'll see what i can do to get this in quickly

mattleibow:
I used this tool: https://github.com/mattleibow/RestSharp.Portable/blob/master/tools/GenAPI.exeTo automatically generate the API. So every time the build runs, I would run it to generate the stubs.

hallem:
@mattleibow where die you get the genapi.exe from?

KraigM:
With a little work, partial PCL support should be possible, which would keep that from being necessary.  Is there anyway to limit the tool to certain files, so the compatible files can stay as they are and the currently incompatible files could slowly transition?

hallem:
@KraigM I'm sure you're correct.  I'm not well versed in the PCL part of the world so if you or @mattleibow would like to take this on and help make it apart of the main library I'd be happy to include it.  I'm hesitant on merging this PR in because of the static-ness of it.  What do you guys say?

KraigM:
Do you say "static-ness" due to the fact that currently the stubs are a snapshot of the api?

hallem:
yes

KraigM:
Fair enough.  I was trying to add something fast that wouldn't effect the rest of your code.  I can see how bad the full conversion would be.  It should be mainly adding an extra condition in a bunch of #if's. 

@hallem Would you prefer #if NETFX_CORE or #if PCL?  Personally I prefer PCL but the default is NETFX_CORE which comes from the original Win8 "Metro" apps.

@mattleibow Let us know any insight on where GenAPI.exe comes from (licensing) and capabilities of limiting scope.

hallem:
@KraigM lets go with #if PCL so there isn't a risk of conflicts anywhere

KraigM:
That was actually less painful then I imagined, though this is not my first time converting a library to PCL compatible. I converted all the current source to be PCL source compatible removing the need for stubs. As stated in the commit, this does not mean that RestSharp has full PCL support. It just means that the API will match, so you will be able to swap in for a supported platform specific version. However, if a few more compatibility issues where handled (mainly Type vs TypeInfo handling), it may very well function (to a point).

KraigM:
Im not sure why the build broke.  It appears it is a unit test that failed, but all I did was pretty much 

#if !SILVERLIGHT && !WINDOWS_PHONE
to
#if !SILVERLIGHT && !WINDOWS_PHONE && !PCL

With a few

#if !PCL

and a few others minor additions for PCL only.  But none of that should have touched the compiled version of Net35 unless it used PCL as a directive.  @hallem any idea?

hallem:
It's not your fault.  There's one test that AppVeyor doesn't like for some reason.

KraigM:
Any way to rerun the build or should I just push a new minor change?

hallem:
no i'll take care of it

hallem:
unless you know whats wrong

KraigM:
No nothing should be wrong

hallem:
i'm going to exclude it from the build for now until i can figure out what's happening, it works locally like a charm

KraigM:
@hallem So whats next?

hallem:
I've got to pull this down and run the tests and merge it in.  I should have time to do it later today or over the weekend.

KraigM:
Let me know if there is anything I need to do

hallem:
there isn't, i'm just super busy sorry

KraigM:
Just trying to get an ETA when this can be merged in?

KraigM:
Implements feature request #760

KraigM:
Is this going to be merged???

davidkaya:
@hallem Hi, any plans for merging this?

CthuluPriest:
Apparently it's never getting merged...

KraigM:
I will make sure this is still working/up-to-date and see about it getting merged

alexsorokoletov:
@KraigM is this pull request abandoned or not?
I know you guys are busy

behoyh:
@KraigM what do you mean by "It just means that the API will match, so you will be able to swap in for a supported platform specific version"?

alexeyzimarev:
@KraigM do you still want PCL or supporting netstandard20 will resolve your issue?

mattleibow:
@hallem @KraigM  Sorry for the late reply...

GenAPI comes from: 
https://www.nuget.org/packages/Microsoft.DotNet.BuildTools.GenAPI/1.0.0-beta-00081

If you add it to the main project, the build will spit out the api automatically.

alexeyzimarev:
We will not support PCL. We go for netstandard instead, the version is to be decided.

mattleibow:
Good point.

For the version, right now, the latest version UWP can support is 1.5. We have to wait for the next release of UWP.
However, you may be able to get away with 1.4. This has a fairly large API coverage.

Of course, 2.0 has a lot more, and waiting may not be a big issue for UWP.

PR #748, opened 2015-09-20
AddUrlSegment overload

mwereda:
AddUrlSegment overload method to allow pass object as value so that
calling ToString() on value is not needed.

Let say one would like to send request to API endpoint `api/document/{documentId}/details` and there is method:
```c#
public IList<DocumentDetails> GetDetails(int documentId)
{
          var request = new RestRequest("api/document/{documentId}/details");
          request.AddUrlSegment("documentId", documentId.ToString());
          ...
}
```
It would be nice not to have to call ToString() method on documentId. Just like this:
```c#
public IList<DocumentDetails> GetDetails(int documentId)
{
          var request = new RestRequest("api/document/{documentId}/details");
          request.AddUrlSegment("documentId", documentId);
          ...
}
```
In my opinion this is more convenient and also thanks to it code is a little bit more cleaner.

### DIFF
--- a/RestSharp.Tests/RestRequestTests.cs
+++ b/RestSharp.Tests/RestRequestTests.cs
@@ -85,8 +85,8 @@ namespace RestSharp.Tests
 
             var parameter = request.Parameters.FirstOrDefault(x => x.Name.Equals(ParameterName));
             Assert.IsNotNull(parameter);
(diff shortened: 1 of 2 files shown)


hallem:
thanks for the PR.  can you also add a unit test for this functionality?

mwereda:
Sure. I've created unit tests for the AddUrlSegment method.

PR #753, opened 2015-09-24
Add support for Matrix Parameters

benfo:
Matrix parameters are another way of defining parameters and can appear anywhere in the path, but before the query string. Matrix parameters are not very common but is supported in the [WADL spec](http://www.w3.org/Submission/wadl/)

Why? Some older intermediaries or proxies don't cache their response with query string parameters and RestSharp does not support it out of the box.

Matrix parameters appear in two forms:
- Non-Boolean parameters are represented as: ';' name '=' value
- Boolean matrix parameters are represented as: ';' name when value is true and are omitted from identifier when value is false

### DIFF
--- a/RestSharp.Tests/UrlBuilderTests.cs
+++ b/RestSharp.Tests/UrlBuilderTests.cs
@@ -188,6 +188,67 @@ namespace RestSharp.Tests
         }
 
         [Test]
+        public void GET_with_matrix_containing_tokens()
+        {
+            RestRequest request = new RestRequest("resource", Method.GET);
+
+            request.AddParameter("foo", "bar", ParameterType.Matrix);
+
+            RestClient client = new RestClient("http://example.com");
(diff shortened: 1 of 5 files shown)


hallem:
thanks for the PR and for including tests.  i've never heard of matrix parameters before, is there a modern use case for them or is this more legacy support?

alexeyzimarev:
Is there any demand for this? It seems like no one uses matrix parameters...

unhammer:
I do! I could put them on manually, but adding them like `QueryParameter`s would be nicer.

alexeyzimarev:
It's been a while...

benfo:
It has been a while...

While it is true that matrix parameters may not be widely supported or used, they can still have some valid use cases. At the time that I submitted this PR, we used it extensible on a project.

That was also the only and last time I personally used it.

PR #764, opened 2015-10-09
Add UnsafeAuthenticatedConnectionSharing flag to RestClient

oleveau:
Add UnsafeAuthenticatedConnectionSharing flag to RestClient so that we
can set it to the underlying webrequest.

### DIFF
--- a/RestSharp/Http.cs
+++ b/RestSharp/Http.cs
@@ -197,6 +197,15 @@ namespace RestSharp
         /// </summary>
         public bool PreAuthenticate { get; set; }
 
+        /// <summary>
+        /// Flag to reuse same connection in the HttpWebRequest
+        /// </summary>
+        public bool UnsafeAuthenticatedConnectionSharing { get; set; }
+
+        /// <summary>
+        /// Flag to send authorisation header with the HttpWebRequest
+        /// </summary>
+        public bool UseUnsafeConnection { get; set; }
 #if FRAMEWORK
         /// <summary>
         /// Proxy info to be sent with request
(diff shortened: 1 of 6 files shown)


hallem:
The changes look good.  Can you fix the merge conflicts and add some unit tests?

RestSharp/Http.cs
```diff
@@ -197,6 +197,15 @@ public Encoding Encoding
...
 
+        /// <summary>
+        /// Flag to reuse same connection in the HttpWebRequest
+        /// </summary>
+        public bool UnsafeAuthenticatedConnectionSharing { get; set; }
+
+        /// <summary>
+        /// Flag to send authorisation header with the HttpWebRequest
+        /// </summary>
+        public bool UseUnsafeConnection { get; set; }
```
KOTRET: unused?

alexeyzimarev: @KOTRET could you elaborate?

KOTRET: @alexeyzimarev prop `UnsafeAuthenticatedConnectionSharing` (Line 203) is definitely in use, but i cannot find usage of the prop `UseUnsafeConnection` (line 208) , which is also added with this commit.

alexeyzimarev: This is a public API in a library. "Unused" in this context makes little sense. Most of the public API methods are "unused" inside the library, they are public because people who download the library will use it.
I agree in a sense that if it is "unused" - it means there are no tests for it. This actually is a problem.

alexeyzimarev: @KOTRET I see what you mean. It is indeed added without any usage.

Issue #742, opened 2015-08-25
Issue uploading files (sync)


wing328:
With the latest version of RestSharp (105.2.2), I ran into an issue uploading a file with the following error message:

```
Error: System.Net.WebException: Request was cancelled. ---> System.IO.IOException: Cannot close the stream until all bytes are written
  --- End of inner exception stack trace ---
  at System.Net.WebConnectionStream.Close () [0x00183] in /private/tmp/source-mono-mac-4.0.0-branch-c5sr3/bockbuild-mono-4.0.0-branch/profiles/mono-mac-xamarin/build-root/mono-4.0.3/mcs/class/System/System.Net/WebConnectionStream.cs:821 
  at System.IO.Stream.Dispose () [0x00000] in /private/tmp/source-mono-mac-4.0.0-branch-c5sr3/bockbuild-mono-4.0.0-branch/profiles/mono-mac-xamarin/build-root/mono-4.0.3/external/referencesource/mscorlib/system/io/stream.cs:260 
  at RestSharp.Http.WriteRequestBody (System.Net.HttpWebRequest webRequest) [0x00000] in <filename unknown>:0 
  at RestSharp.Http.PostPutInternal (System.String method) [0x00000] in <filename unknown>:0 
  at RestSharp.Http.AsPost (System.String httpMethod) [0x00000] in <filename unknown>:0 
  at RestSharp.RestClient.DoExecuteAsPost (IHttp http, System.String method) [0x00000] in <filename unknown>:0 
  at RestSharp.RestClient.Execute (IRestRequest request, System.String httpMethod, System.Func`3 getResponse) [0x00000] in <filename unknown>:0 
```

Using an older version of RestSharp (105.1.0), I was able to upload the file without issue. Here is the code to repeat the issue.

```
using System;
using RestSharp;
using RestSharp.Extensions;
using System.IO;

namespace UploadTest
{
    class MainClass
    {
        public static void Main (string[] args)
        {
            var client = new RestClient("http://petstore.swagger.io/v2/");
            var request = new RestRequest("pet/{petId}/uploadImage", Method.POST);
            request.AddUrlSegment("petId", "2"); // replaces matching token in request.Resource

            FileStream stream = new FileStream ("/var/tmp/1.txt", FileMode.Open);

            var fileParam = FileParameter.Create("file", stream.ReadAsBytes(), Path.GetFileName(((FileStream)stream).Name));
[...]

PR #769, opened 2015-10-22
Fix content length of file parameter

dgreenbean:
Adding required input of content length when adding a file using a writer action.

### DIFF
--- a/RestSharp/IRestRequest.cs
+++ b/RestSharp/IRestRequest.cs
@@ -158,9 +158,10 @@ namespace RestSharp
         /// <param name="name">The parameter name to use in the request</param>
         /// <param name="writer">A function that writes directly to the stream.  Should NOT close the stream.</param>
         /// <param name="fileName">The file name to use for the uploaded file</param>
+        /// <param name="contentLength">The length (in bytes) of the file content.</param>
         /// <param name="contentType">The MIME type of the file to upload</param>
         /// <returns>This request</returns>
-        IRestRequest AddFile(string name, Action<Stream> writer, string fileName, string contentType = null);
+        IRestRequest AddFile(string name, Action<Stream> writer, string fileName, long contentLength, string contentType = null);
 
         /// <summary>
         /// Add bytes to the Files collection as if it was a file of specific type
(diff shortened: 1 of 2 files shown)


dgreenbean:
This is to address issue https://github.com/restsharp/RestSharp/issues/742.

RestSharp/IRestRequest.cs
```diff
@@ -158,9 +158,10 @@ public interface IRestRequest
         /// <param name="name">The parameter name to use in the request</param>
         /// <param name="writer">A function that writes directly to the stream.  Should NOT close the stream.</param>
         /// <param name="fileName">The file name to use for the uploaded file</param>
+	/// <param name="contentLength">The length (in bytes) of the file content.</param>
```
hallem: Please fix the spacing difference here.

RestSharp/RestRequest.cs
```diff
@@ -172,15 +172,17 @@ public IRestRequest AddFile(string name, byte[] bytes, string fileName, string c
         /// <param name="name">The parameter name to use in the request</param>
         /// <param name="writer">A function that writes directly to the stream.  Should NOT close the stream.</param>
         /// <param name="fileName">The file name to use for the uploaded file</param>
+	/// <param name="contentLength">The length (in bytes) of the file content.</param>
```
hallem: Please fix the spacing difference here.

hallem:
Code looks good. Just some minor things that need to be changed first.  Also, any idea on why this broke?  I haven't looked into it too deeply yet but I'm concerned about changing the method signature and breaking other implementations.

dgreenbean:
I didn't look too closely at why this broke.  Since the method is currently broken anyway, I wasn't too concerned about changing the signature.  If you think the compiler errors could cause problems for people, I could make this method throw an Exception and add a new method that works.  That felt like overkill to me.

dgreenbean:
The spacing should be fixed now.  I'm guessing at the moment that the AppVeyor failure is unrelated to the change.  Please let me know if this is not the case.

hallem:
it is unrelated unless it doesn't run successfully locally :)

dgreenbean:
@hallem Is there anything blocking this from being merged?

hallem:
no i just haven't had time to sit down and work on this.  i've recently moved so i'm still dealing with all that.  i should have some time soon.

PR #773, opened 2015-10-29
Detection of millisecond-timestamps

Falco20019:
ParseJsonDate detects if the given long is outside the allowed range and tries to add it as milliseconds instead of seconds.

This was a problem for me when using Dates generated with Java (Date.getTime()).

### DIFF
--- a/RestSharp.Tests/JsonTests.cs
+++ b/RestSharp.Tests/JsonTests.cs
@@ -624,6 +624,17 @@ namespace RestSharp.Tests
         }
 
         [Test]
+        public void Can_Deserialize_Unix_Json_Millisecond_Dates()
+        {
+            string doc = CreateUnixDateMillisecondsJson();
+            JsonDeserializer d = new JsonDeserializer();
+            RestResponse response = new RestResponse { Content = doc };
+            Birthdate bd = d.Deserialize<Birthdate>(response);
+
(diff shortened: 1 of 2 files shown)


Falco20019:
RestSharp.IntegrationTests.AsyncTests.Can_Cancel_GET_TaskAsync fails, but nothing I have touched :)

hallem:
The test failure is a problem with the CI build on AppVeyor.  I've not been able to figure out why it's failing.

As for the PR, great idea and thanks for the included tests.  Have you considered the broader impacts on this change?  It may cause other things to break.  Can you elaborate more on what the problem you're facing is?

Falco20019:
It shouldn't break anything, since the valid timestamps in milliseconds (or at least all relevant ones) are higher than the allowed values DateTime can take. The last ambigous timestamp where it can't tell apart if it's milliseconds or seconds is 253402300 (Wed, 11 Jan 1978 21:31:40 GMT). And for these cases, I left the original behaviour of using seconds. The first valid timestamp for milliseconds I assume is 253402300800, which would be the Sat, 1 Jan 10000 00:00:00 GMT if calculated in seconds.

The problem is, that I use [Date.getTime()](http://docs.oracle.com/javase/7/docs/api/java/util/Date.html#getTime%28%29) of Java to generate my timestamp. These are in milliseconds and RestEasy of JBoss does not add "new Date(...)" or "Date(...)" as assumed here. So RestSharp has to make a more flexible assumption here.

PR #795, opened 2015-12-18
Added ProtocolVersion to RestResponse.

### LINKED ISSUE #790: Protocol Version is not available
I would like to use the HTTP version (1.0, 1.1, etc) from a `RestResponse`. This isn't available on `RestResponse` but it is available on the underlying `HttpWebResponse` as the `ProtocolVersion` property. Was this left out intentionally?

jeffesp:
It is only set in FRAMEWORK builds, but is available as the default on any platform.

Fixes #790


jeffesp:
I had an earlier PR for #790, but wouldn't pass builds. So I went back at it with this one and I think it should be all clear.

hallem:
That one test has been problematic and I've not been able to figure out why.

PR #801, opened 2016-01-14
Fix XmlDeserializer for case when RootElement depth >1

### LINKED ISSUE #800: XmlDeserializer doesn't process case when RootElement has depth > 1
There is only one test case with RootElement, and it covers only depth = 1 case (please see it here: https://github.com/restsharp/RestSharp/blob/master/RestSharp.Tests/XmlDeserializerTests.cs#L103 )

When I'm saying 'depth' here I mean depth in XML hierarchy, i.e. for XML <root><subroot><subsubroot></subsubroot></subroot>/<root> subroot depth = 1, while subsubroot depth = 2.

ma-gu:
Closes #800.

### DIFF
--- a/RestSharp.Tests/XmlDeserializerTests.cs
+++ b/RestSharp.Tests/XmlDeserializerTests.cs
@@ -81,6 +81,19 @@ namespace RestSharp.Tests
         }
 
         [Test]
+        public void Can_Deserialize_When_RootElement_Deeper_Then_One()
+        {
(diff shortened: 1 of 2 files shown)


ma-gu:
Locally all tests are green (excluding 5 skipped which require consumer key/secret).

uxsoft:
@hallem I have tested this and I can confirm the changes work as expected. :shipit:

PR #802, opened 2016-01-15
main

StefH:


hallem:
@StefH not sure what is up with this PR but there are two commits and no files changed.  did something happen?

StefH:
I did probably delete my branch because nothing happened for 10 months. I will try to check my local system if I can find my changes.

Issue #820, opened 2016-03-02
DownloadData hides errors when download fails


robi26:
When using RestClient.DownloadData to download a file and the file does not exist or the server returns an error the exception is hidden by the DownloadData method and cannot be handled.

alexeyzimarev:
Now you can use the new overload, which accepts the `throwOnError` parameter.

davetoland:
This doesn't appear to be working for me, I have set throwOnError to true, and even tried adding a line above to tell the client to ThrowOnAnyError, and whilst Visual Studio reports a 400 exception, the DownloadData method still swallows it and continues...

`try
{
    var client = GetRestClient(baseUri, credential);
    client.ThrowOnAnyError = true;
    var data = client.DownloadData(new RestRequest(downloadLink), true);
    File.WriteAllBytes(localPath, data); //<-- still gets executed
    result = true;
}
catch (Exception e)
{
    Logger.Error($"Error downloading Attachment from link: {downloadLink}", e);
}`

davetoland:
![image](https://user-images.githubusercontent.com/19878260/75982775-3c438f80-5edf-11ea-9902-365332283b06.png)

![image](https://user-images.githubusercontent.com/19878260/75982784-41a0da00-5edf-11ea-8130-6a764260f046.png)

davetoland:
Ok, so I've grabbed your source and built in in Debug to try and work out what's wrong here. Seems you're only throwing if the response is not an HttpWebResponse. So... by design. 
The situation for my test, is that we're intentionally passing in an incomplete URL, it targets a directory above the intended target (i.e. "image.png" fragment is missing from the end of the URL) and we're expecting the client to raise the exception.
As things stand, we've now got a byte array returned which File.WriteAllBytes happilly saves to disk, but we've got no way of telling that the file is not the intended target. 
If I try to open the file (it's an image), Windows reports a problem, that it can't open it, as you'd expect... but obviously that doesn't get caught in the code. 
Any suggestions?

davetoland:
![image](https://user-images.githubusercontent.com/19878260/75984740-deb14200-5ee2-11ea-9ed8-65223a1dbc79.png)

davetoland:
[...]

PR #821, opened 2016-03-02
Throw an exception if the download fails

robi26:
Fixes issue #820

### DIFF
--- a/RestSharp/RestClient.Sync.cs
+++ b/RestSharp/RestClient.Sync.cs
@@ -14,6 +14,10 @@ namespace RestSharp
         public byte[] DownloadData(IRestRequest request)
         {
             IRestResponse response = this.Execute(request);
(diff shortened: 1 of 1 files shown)


hallem:
This would be a breaking change for some people.  Is it possible for you to do this in a way that preserves the default behavior and makes this an option?  Or do you have a convincing argument on why we should make this breaking change?

robi26:
My argument for throwing exceptions is that you can get invalid data. Assume you download a .pdf and get a 404 error page. The now downloaded bytes contain the error html but the file is still a .pdf. How can I ever detect, if I downloaded a .pdf or got a 404?

We could introduce a configuration variable 'CatchExceptions' which is false by default and can be set to true to get the exceptions.

Please let me know what you think.

alexeyzimarev:
I suggest we return null instead. Alternatively, we can have another method.

PR #824, opened 2016-03-08
Implement valid JSON serialization and deserialization for NaN, -INF and +INF.

mschwerin:


hallem:
We can't really make changes to that file.  It's something we take a dependency on.

mschwerin:
Hi,

sorry I missed that. You might want to look at

https://github.com/facebook-csharp-sdk/simple-json/pull/69/commits/c1b769825deb89f4bf628a223ec866a8e6081a73

They seem to have addressed the problem already...

Cheers, m.

PR #836, opened 2016-04-20
Supports PATCH - update readme

geek:
https://github.com/restsharp/RestSharp/blob/master/RestSharp/RestClient.Sync.cs#L34


uxsoft:
@hallem This can obviously be accepted :shipit:

PR #841, opened 2016-05-15
User set Content-Type was being by the content type of the body

### LINKED ISSUE #784: Cannot set ContentType = "application/x-www-form-urlencoded; charset=UTF-8"
```
var request = new RestRequest(Method.POST);
request.AddParameter("nummessage", 1);
request.AddParameter("mediatype", 31);

var client = new RestClient("LemonService.php") { Encoding = Encoding.UTF8 };
var response = client.Execute(request);
```

My `LemonService.php` expected that RestSharp will send the request with the following header

`content-type: application/x-www-form-urlencoded; charset=UTF-8`

but RestSharp only send

`content-type: application/x-www-form-urlencoded`

There is nothing in the request that tell my `LemonService.php` that the request encoding was UTF-8.

I will have to come back to `System.Net.WebRequest` for work-around

```
WebRequest webRequest = WebRequest.Create("LemonService.php"");
webRequest.ContentType = "application/x-www-form-urlencoded; charset=UTF-8";
webRequest.Method = "POST";
String request = "nummessage=" + System.Web.HttpUtility.UrlEncode("1", Encoding.UTF8)
"&mediatype=" + HttpUtility.UrlEncode("31", Encoding.UTF8)
byte[] bytes = Encoding.ASCII.GetBytes(request);
webRequest.ContentLength = bytes.Length;   //Count bytes to send
using (var os = webRequest.GetRequestStream()) {...}
```

uxsoft:
Fixes #784 where users were unable to set a custom content type because it was being overwritten by the content type of the body.

### DIFF
--- a/RestSharp.IntegrationTests/StatusCodeTests.cs
+++ b/RestSharp.IntegrationTests/StatusCodeTests.cs
@@ -56,9 +56,9 @@ namespace RestSharp.IntegrationTests
             {
                 RestClient client = new RestClient(baseUrl);
                 RestRequest request = new RestRequest("error")
-                                      {
-                                          RootElement = "Success"
-                                      };
+                {
+                    RootElement = "Success"
+                };
 
                 request.OnBeforeDeserialization = resp =>
                                                   {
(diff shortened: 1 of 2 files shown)


RestSharp.IntegrationTests/StatusCodeTests.cs
```diff
@@ -102,10 +102,40 @@ public void Handles_Default_Root_Element_On_No_Error()
                 Assert.AreEqual("Works!", response.Data.Message);
             }
         }
+
+        [Test]
+        public void ContentType_Additional_Information()
```
mjwsteenbergen: It would be great if we could have tests for all of the if statements.

mjwsteenbergen:
Thank you for your contribution! I see your pull request has some merge conflicts. To be considered for merging, please merge the latest changes into your branch.

PR #842, opened 2016-05-15
Fixed SHA256 for UWP

uxsoft:
Fixes a recent addition of SHA256 to build on UWP.
Added a compiler directive exactly like it is when handling SHA1.


alexeyzimarev:
I will merge the changes manually, the formatting is broken, I will take the code.

PR #843, opened 2016-05-17
query string parameters are encoded using the currently selected encoding instead of default encoding

masadwin:
No matter what encoding I selected, the Url query string parameters were
always encoded using the default UTFencoding.
Reason: i needed to call a GET method that required me to send my query
string encoded in ISO-8859-1 encoding. I tried many options but 'ø'
would always become '%C3%B8' even though I had set the encoding property
to ISO-8859-1 which should get me '%F8'
This change uses the currently selected encoding in the RestClient.cs
and uses it to encode parameters and names instead of using UrlEncode
which (although works for most cases) might not work for everybody.

### DIFF
--- a/RestSharp/RestClient.cs
+++ b/RestSharp/RestClient.cs
@@ -344,6 +344,7 @@ namespace RestSharp
             Authenticator?.Authenticate(client, request);
 
         private static string EncodeParameters(IEnumerable<Parameter> parameters, Encoding encoding) =>
+            string.Join("&", parameters.Select(parameter => EncodeParameter(parameter, encoding)).ToArray());
 
         private static string EncodeParameter(Parameter parameter, Encoding encoding) =>
             parameter.Value == null


alexeyzimarev:
I don't think it is a good idea to reference MonoHttp namespace in shared code. Can you comeup with some alternative? I understand your issue. May be start with a new unit test?

masadwin:
sure, let me make the necessary changes

alexeyzimarev:
@masadwin How is it going? Also, remember to rebase to the latest `develop`, now we have conflicts.

masadwin:
hello @alexeyzimarev
I am sorry I have not been able to take a look at this after my last update. But thanks for reminding me. I will get to it asap. Will take a look at the alternative if MonoHttp is not desired along with a test case for it :)

masadwin:
closing this pull request and opening another because i kind of messed this repository up locally

masadwin:
https://github.com/restsharp/RestSharp/pull/1034
here

PR #850, opened 2016-05-24
AddOrUpdateParameter, convenience method

andrewchaa:
I had a case where I had to check if some parameters already exist and replace the values, and if they don't, add those value as new. I think It'll be nice if we have this convenience method in the library.


alexeyzimarev:
Would be nice if you can complete your code: https://github.com/restsharp/RestSharp/issues/1002

andrewchaa:
No problem. Will do.
Thanks for letting me know.

Andrew.

PR #862, opened 2016-07-01
Compile with .NET Core

zhongzf:


zhongzf:
Add compile condition:
#if NETCORE50 || NETSTANDARD1_5 || NETSTANDARD1_6

zhongzf:
Test nuget package can be found here:
https://www.nuget.org/packages/RestSharp.NetCore/105.2.4-rc4-24214-01

alexeyzimarev:
There is no need to use `ifdef` where we refactor `System.Type` to `System.TypeInfo`. `System.TypeInfo` always works. We should have been just use that, unconditionally.

Anyway, we will not have a separate package, but use multi-framework targeting in VS2017 project format.

PR #873, opened 2016-09-07
Eliminate spurious ToString calls during deserialization

logiclrd:
I noticed some time ago that the `ConvertValue` method in `JsonDeserializer` always `.ToString()`s the object before checking what kind of deserialization must be performed. Most of the _types_ of deserialization do require the string, but the most _frequent_ types of deserialization will not. This becomes pathological in the case of a deep, complex, typed object structure, because each layer gets passed through `ConvertValue` separately. In this PR, I've reworked the `ConvertValue` method to avoid the `ToString` call unless it is strictly required. I added a custom-built `Lazy`-type class, since as I understand it `Lazy<T>` does not exist in .NET 3.5. I also added a unit test that ensures the offending `ToString` calls are not made. (For what it's worth, the `ToStringCounter` value on the original `ConvertValue` code has value 121 at the end of the test!)

PR #874, opened 2016-09-07
Allow deserialization of very large integers

logiclrd:
While I was working on PR #873, I was reading through the SimpleJson code and discovered that the way in which it parses integers might reject some syntactically-valid JSON documents. In particular, JSON allows for very large integer values to be specified (values large enough that a 64-bit integer cannot represent them) with no punctuation or exponentiation. These values would, in typical implementations, need to be represented as floating-point values (`double`), but because they lack periods and `e` signifiers, the SimpleJson code will only attempt to parse them as `long`. This means that the (valid) JSON string produced by `(long.MaxValue + 1M).ToString()` cannot be parsed. This PR addresses the issue (using `decimal` and then `double` as fallbacks where `long` fails), and adds a corresponding unit test.

### DIFF
--- a/RestSharp.Tests/JsonTests.cs
+++ b/RestSharp.Tests/JsonTests.cs
@@ -825,6 +825,14 @@ namespace RestSharp.Tests
             Assert.IsNull(dictionary["Null"]);
         }
 
+        [Test]
+        public void Can_Deserialize_Very_Large_Integers()
+        {
+            decimal veryLargeInteger = long.MaxValue + 1M;
+            decimal deserializedValue = new JsonDeserializer().Deserialize<decimal>(new RestResponse { Content = veryLargeInteger.ToString() });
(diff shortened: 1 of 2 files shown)


behoyh:
SimpleJson is a dependency, we can't modify it.

logiclrd:
Ahh, okay. Where does it live? I can try pushing this patch upstream. :-)

logiclrd:
I believe I have found it: https://github.com/facebook-csharp-sdk/simple-json

I'll create a PR there. Looks like `SimpleJson.cs` is brought in via NuGet? Kind of strange, but I assume that means if my PR is accepted, then I can come back to RestSharp and create a PR consisting of updating the NuGet reference?

logiclrd:
https://github.com/facebook-csharp-sdk/simple-json/pull/75

PR #878, opened 2016-09-21
Change restricted header action for If-Modified-Since to use Invariant CultureInfo

mrknmc:
We were having issues communicating between services using different calendar formats, because the date restriction in the header action was not accepting a gregorian calendar date from the caller as its culture was set to th-TH (using a buddhist calendar), but the receiving service would not have accepted a th-TH style date. 
The fix is to change the date formatting of the `If-Modified-Since` header to use the invariant CultureInfo. This allows services to use RFC1123 formatting for the `If-Modified-Since` header to get around any localization issues.

For example: currently, with the current culture of the service set to th-TH and the `If-Modified-Since` header equal to `"Thu, 15 Sep 2016 20:07:00 GMT"` the Execute method returns an error response with the following trace:

```
   at System.DateTimeParse.Parse(String s, DateTimeFormatInfo dtfi, DateTimeStyles styles)
   at System.Convert.ToDateTime(String value)
   at RestSharp.Http.<AddSyncHeaderActions>b__25(HttpWebRequest r, String v)
   at RestSharp.Http.AppendHeaders(HttpWebRequest webRequest)
   at RestSharp.Http.ConfigureWebRequest(String method, Uri url)
   at RestSharp.Http.GetStyleMethodInternal(String method)
   at RestSharp.RestClient.Execute(IRestRequest request, String httpMethod, Func`3 getResponse)
```

With this fix in place, Execute allows the header to be in the RFC1123 format no matter what the current culture of the service is.


alexeyzimarev:
Can you add a unit test for that use case you have seen previously failing and now fixed by your code?

PR #891, opened 2016-10-18
Fixing file lock from BuildTask

edgarrc:
Chances that you haven't seen this error before because the use of CI servers.

If you manually build/clean/rebuild on VS, the RestSharp.Build.dll will get called twice on the same VS memory context. It will throw an error: 

> "Severity Code    Description Project File    Line    Suppression State
> Error     Unable to copy file [....dll] to [....dll]. The process cannot access the file '..dll' because it is being used by another process."

This happens because Assembly.LoadFile(path) locks the file. I changed to File.ReadAllBytes, so the file will no longer get locked.

If this is accepted, .nuget\RestSharp.Build.dll should be recompiled.


alexeyzimarev:
Build task is no longer used.

PR #897, opened 2016-11-10
changed Regex suffix to static

Bobs13:
This is perfomance problem. When we created class RestClient, it is checked AddHandler, but we create new regexp every time. And this is very long operation. If use static, this operation very fast.
Example:
            for (var i = 0; i < 1000; i++)
            {
                var client = new RestSharp.RestClient( new Uri("https://api-maps.yandex.ru/2.1/"));
            }
with static - 0.07s
without static - 5s
msdn is written regex is Thread Safety - https://msdn.microsoft.com/en-us/library/6h453d2h(v=vs.110).aspx

### DIFF
--- a/RestSharp/RestClient.cs
+++ b/RestSharp/RestClient.cs
@@ -201,7 +201,7 @@ namespace RestSharp
         {
             this.ContentHandlers[contentType] = deserializer;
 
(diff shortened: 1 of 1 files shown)


IanBuswell-etech:
I'd like to see this merged in please.

Issue #905, opened 2016-12-07
RestClient.DefaultParameters allows only single Querystring parameter per key


Krizzzn:
The RestClient drops Parameters when they already exist. This might be okay for Headers, but Form and Querystring Parameters should be allowed multiple times. 

See `RestClient.cs` 410 and following.

Krizzzn:
Will provide a PR for this tomorrow.

This change might have impact on existing code bases as it changes the behaviour of the RestClient. Even though the new behaviour is more correct, it's debatable if the PR should be accepted.

PR #906, opened 2016-12-08
Allowing multiple Default Parameters for same key

Krizzzn:
Fixes issue #905


Krizzzn:
This change might have impact on existing code bases as it changes the behaviour of the RestClient. Even though the new behaviour is correct, it's debatable if the PR should be accepted.

alexeyzimarev:
I added a property to keep the default behaviour (not allowing multiple parameters with the same name). If the property is set to `true` - query and form default parameters with the same name will be allowed. Needs tests, I forgot :(

Issue #760, opened 2015-09-30
Issue with trying to install RestSharp for a Portable Class through Nuget


jamesbar2:
Could not install package 'RestSharp 105.2.3'. You are trying to install this package into a project that targets 'portable-net45+win+MonoAndroid10+MonoTouch10', but the package does not contain any assembly references or content files that are compatible with that framework. For more information, contact the package author.

![image](https://cloud.githubusercontent.com/assets/828489/10209470/3bda9bcc-67a1-11e5-83ab-15fb68f24560.png)

Let me know if there's a workaround for this, or what can be done. Thanks!

ravensorb:
Unfortunately as far as I know RestSharp is not currently compatible with PCL -- I am hoping this will change soon though :smile:

JeansReal:
+1

hallem:
the library is not able to target multiple PCLs.  It can target iOS, Andriod, etc. individually but not all at once yet.

KraigM:
It is supported under #734 once it gets merged in

jamesbar2:
Awesome! Thank you!

KraigM:
Currently it is PCL Profile78 (aka .Net 4.5+, Xamarin.iOS, Xamarin.Android, Xamarin.Mac, WP App 8+, Windows 8+, WP Silverlight 8+, Windows Store App).  

The PCL version should mostly work but each of those platforms should already be supported with platform specific versions so in the final application NuGet will choose the platform specific version (which is what you want).  But you will be able to compile a full PCL (Profile 78 or compatibile) library using RestSharp.PCL and it will be compatible with the platform version of RestSharp at runtime (JIT or AOT).

sunthx:
+1

xinmeng1:
+1

aniketghode:
Actually, you can add RestSharp package to either iOS or Droid project in your solution and then in PCL browse to your solutionName/packages/RestSharp to find a RestSharp.dll and add it as a reference in your PCL project. i

It should work.

gentledepp:
Hey guys - I just finished the PCL support. But until it is merged, you'll have to clone the repo yourselves and run the build.bat
Happy coding ;-)

alexeyzimarev:
Sorry, this is very obsolete. .NET Standard 2.0 or .NET Framework 4.5.2 only now.

PR #908, opened 2016-12-09
PCL using bait and switch trick & finished UWP support

gentledepp:
Hi!
I took the liberty to finish the (nearly ready) UWP implementation as we need it. (I think there was only a missing HMAC256 signature)

Also, using the bait-and-switch trick, I added PCL support.  (see: http://log.paulbetts.org/the-bait-and-switch-pcl-trick/) which closes issue #760
The build.bat is also adapted to create the nuget packages with PCL folder.

Build runs fine, but this is **not tested yet!**

I'll keep you in the loop though :)


gentledepp:
Hey guys - I need your assistance, please!
1) the appveyor build fails. It complains that the RestSharp.UWP project has no outputdir specified for Configuration='Debug.Appveyor'  Platform='AnyCPU'
however, locally, when switching to that configuration, the build works like a charm.
I don't get it... :-|

2) One particular test is failing: InterfaceImplementationTests.IRestSharp_Has_All_RestSharp_Signatures
The reason might be, that I refactored the IRestClient interface:
I pulled all synchronous methods into a IRestClientSync interface, 
      and all asynchronous methods into a IRestClientAsync interface.
The reason is, that I want to implement the interface in a RestClientWrapper class in my PCL project, where no synchronous "Execute" methods exist.
Now, when I ran the unittests, the .NET45 version of RestSharp - which has those synchronous methods in the IRestClient interface - was pulled in and I got a MissingMethodException thrown at me at runtime because "method Execute is not implemented in RestClientWrapper". 
=>So I had to divide those interfaces. 
(I hope my argumentation is clear? :-|)

Other than that - the bait-and-switch version works like a charm!
**NOTE**
* if you are still using nuget with package.config, you have to reference RestSharp both in your PCL and your platform (e.g. Xamarin.Android) project in order to pull-in the platform specific RestSharp version
* if you are using nuget with project.json already, you only have to reference RestSharp in your PCL, as then nuget automagically pulls-in the RestSharp dll which suits the current platform best!

alexeyzimarev:
I am sorry but I think we will go towards .NET Standard instead of PCL for UWP and all other platforms.

PR #915, opened 2016-12-24
Fix RestSharp.UWP build for UAP

zabulus:
Assembly produced after the fix, can be consumed in uap10.0 projects.


medianick:
This PR appears to fix what is breaking all the other recent PRs; the AppVeyor build failure with a missing OutputPath in the RestSharp.UWP.csproj project. Please merge!

zabulus:
According to this:
https://gist.github.com/davidfowl/8939f305567e1755412d6dc0b8baf1b7
The change should slightly other. There should be one library that targets lowest available netstandart. And all other targets should be removed at all.
Still the PR addresses the problem I experienced with using RestSharp

alexeyzimarev:
UWP builds now. Thank you for your effort.

PR #919, opened 2017-01-02
Credentials property moved to Client from Request.

ertant:
Like ClientCertificates property should be on Client not Request. Internal Credentials class automatically fetch the credentials by accessed resource URL.


alexeyzimarev:
May be moving it was not so good idea? Can't we have it in both places? Like by default we take if from the client but you can override credentials in request? It will solve your issue and also keep the API compatible,

PR #931, opened 2017-02-16
Clicking the image opens the image in a new tab

weitzhandler:
Should refer to the location you want the users to look at.


alexeyzimarev:
I actually removed this. Assuming we got enough people to move on, plus the author of that is not talking to people anyway now.

mjwsteenbergen:
True

Issue #940, opened 2017-03-28
How  to write c# class     serialize like this xml  ?


maikebing:
```
 <?xml version="1.0" encoding="UTF-8"?>
    <Messages xmlns="http://mns.aliyuncs.com/doc/v1/">
        <Message>
            <MessageBody>Message1 Base64 Encoded Result</MessageBody>
            <DelaySeconds>60</DelaySeconds>
            <Priority>1</Priority>
        </Message>
        <Message>
            <MessageBody>Message2 Base64 Encoded Result</MessageBody>
            <DelaySeconds>60</DelaySeconds>
            <Priority>1</Priority>
        </Message>
    </Messages>
```

alexeyzimarev:
You can easily use the schema generator tool to produce csharp files from XML.
https://blogs.msdn.microsoft.com/yojoshi/2011/05/14/auto-generating-entity-classes-with-xsd-exe-for-xml-serialization-and-de-serialization/

PR #943, opened 2017-03-31
XmlElementAttribute Support!

maikebing:
Indicates that a public field or property represents an XML element when the
 System.Xml.Serialization.XmlSerializer serializes or deserializes the object
that contains it.  #940


maikebing:
@johnsheehan  can you merge this ?

alexeyzimarev:
This needs work. I will try to fix it.

Issue #963, opened 2017-06-27
Xml node with attributes and content at the same time


bcichonski:
It seems to me that when using default XmlSerializer there is no way to produce xml node that has attribute values and text content at the same time. This was originally found when my team was using RestSharp for communicating with Prestashop 1.6 API, where some nodes should look like:

```xml
<language id="1">
 Beer
</language>
```

PS: I ommit Google Group since it looks dead and abandoned.

alexeyzimarev:
Closed by #964

PR #964, opened 2017-06-27
Xml node with attributes and content at the same time

bcichonski:
#963 proposed implementation

### DIFF
--- a/RestSharp.Tests/SampleClasses/misc.cs
+++ b/RestSharp.Tests/SampleClasses/misc.cs
@@ -19,6 +19,7 @@
 using System;
 using System.Collections.Generic;
 using RestSharp.Serializers;
+using RestSharp.Deserializers;
 
 namespace RestSharp.Tests.SampleClasses
 {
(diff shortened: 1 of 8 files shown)


alexeyzimarev:
Please rebase and squash.

bcichonski:
I see commits have been squashed already - thanks. I have looked (briefly) at those conflicts and it seems to me that they could be resolved mostly by accepting my changes or all the methods (in tests), unfortunately I don't have write access to do so. Please let me know if I could help in any way.

bcichonski:
I see commits have been squashed already - thanks. I have looked (briefly) at those conflicts and it seems to me that they could be resolved mostly by accepting my changes or all the methods (in tests), unfortunately I do not have write access to do so. Please let me know if i could help in any way.

alexeyzimarev:
You can rebase, you only need access to your own branch to do so. You can just follow these instructions: https://github.com/edx/edx-platform/wiki/How-to-Rebase-a-Pull-Request

Please remember to review the diffs for any merge because some stuff that was changed is important. If you encounter using `var` where you use explicit type - this is OK, you can keep your code as-is.

bcichonski:
I did squashing and rebasing, but now 19 deserialization tests fail. As far as I remember, it was not so in June. Worse still, I have some strange problems when debugging these tests. I can not access the source library of RestSharp from the test library, what also worked well before. I need more time to investigate.

Issue #867, opened 2016-07-29
Encoding set problem


shanvenleo:
HttpResponse.cs
public string Content
 {
            get { return this.content ?? (this.content = this.RawBytes.AsString()); }
 }

when i set 
`RestClient.Encoding = Encoding.GetEncoding("gb2312");`
this method called AsString() and use Encoding.UTF8,not use the RestClient.Encoding value

an alternative 
`string content = Encoding.GetEncoding("gb2312").GetString(response.RawBytes);`

maobiwen:
看样子又是一个中国的开发者啊。
尼玛感觉我也被坑了。调试了好久。。。。

alexeyzimarev:
We cannot guarantee that the response encoding matches the client encoding. However, we will try parsing the response encoding and fall back to UTF8 if we can't.

ueberboss:
The current fix causes first chance exceptions when a Content-Encoding is not specified. See: https://stackoverflow.com/questions/55052160/restsharp-first-chance-exception-am-i-using-it-wrong

A possible fix for this:

```
public static string AsString(this byte[] buffer, string encoding)
{
	Encoding enc;
	if (string.IsNullOrEmpty(encoding)) // avoid an unnecessary exception
	{
		enc = Encoding.UTF8;
	}
	else
	{
		try
		{
			enc = Encoding.GetEncoding(encoding);
		}
		catch (Exception)
		{
			enc = Encoding.UTF8;
		}
	}

	return AsString(buffer, enc);
}
```

in \RestSharp-master\RestSharp\Extensions\MiscExtensions.cs

Robelind:
This issue needs to be re-opened.

kealist:
@alexeyzimarev  I am getting exceptions when content encoding is not specified as well

carlhoogstoel:
I have to agree with the OP. The AsString method in the MiscExtensions.cs needs to be fixed.

MichelMichels:
I get the same exceptions because my API provider can't properly send the content-type in the response.

jiangfeng0103:
> 看样子又是一个中国的开发者啊。
> 尼玛感觉我也被坑了。调试了好久。。。。

这个问题解决了吗？现在还是有这个问题

maobiwen:
> > 看样子又是一个中国的开发者啊。
> > 尼玛感觉我也被坑了。调试了好久。。。。
> 
> 这个问题解决了吗？现在还是有这个问题

[...]

PR #972, opened 2017-07-21
Use Encoding to UrlEncode the post parameters 

linjiacheng:
Use the assigned Encoding to UrlEncode the values of post parameters.

It always fails when post non utf-8 values to a not utf-8 page.

### DIFF
--- a/RestSharp/Http.cs
+++ b/RestSharp/Http.cs
@@ -24,10 +24,6 @@ using System.Net;
 using System.Text;
 using RestSharp.Extensions;
 
-#if WINDOWS_PHONE
-using System.Net;
-#endif
-
 #if SILVERLIGHT
 using System.Windows.Browser;
 #endif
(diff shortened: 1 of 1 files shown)


RestSharp/Http.cs
```diff
@@ -360,7 +363,13 @@ private string EncodeParameters()
                     querystring.Append("&");
                 }
 
-                querystring.AppendFormat("{0}={1}", p.Name.UrlEncode(), p.Value.UrlEncode());
+                querystring.AppendFormat("{0}={1}", p.Name.UrlEncode(),
```
mjwsteenbergen: Why do you want to use the `HttpUtility` encode the value, but not the name?

alexeyzimarev:
@mjwsteenbergen there is an issue https://github.com/restsharp/RestSharp/issues/867

alexeyzimarev:
There is no update and there is no test confirming the issue, also using custom-made `HttpUtils` is not the best idea, especially considering them to be removed by now.

Feel free to resubmit, to develop, with the test.

PR #977, opened 2017-08-07
Update Http.cs

JxiaoC:
2G File ,bad


mjwsteenbergen:
Can you elaborate why this change is necessary?

alexeyzimarev:
Duplicate of #936. We cannot merge it since it fails the test. Test fails because they build and run on .NET 3.5. We will merge #936 when we move the test project to newer framework.

PR #1004, opened 2017-09-12
Ported the code to Windows Compact Framework

kendallb:
## Description

We have a number of LOB applications that currently run on Windows Mobile devices using Compact Framework 3.5. It's a terrible platform to develop for, but we still use it because it makes it easy to have relatively portable code to go between Windows Forms on the desktop and Windows Forms on Compact Framework. Anyway long story short we are in the process of porting all our code to run on Android devices and also HTML5 browser apps instead (same device, just running Android) and as part of that process to simplify that process we are re-writing all our WCF services in REST so they are easier to use on any platform especially from Javascript in the browser. So as a stepping stone we needed support for RestSharp on Compact Framework as we use that for all our REST service clients. 

This change is the minimal set of changes to port to Compact Framework. Most of it revolves our language features missing in the version of C# in Visual Studio 2008, the lack of Cookies on Compact Framework and some stuff that simply does not work. So far it has been testing out to work just great for us, but I will be doing more extensive testing before we deploy to production in the next week.

Most of the changes are relatively innocuous but some stuff like the lack of optional parameters are just plain annoying as it makes the code messier for all platforms. So for that reason if there is no interest in merging this change set into the mainline tree, I would be happy to build a new repository for a Compact Framework version and make it a pure fork. Toss all the stuff that is not needed from the project and just include the Compact Framework support and we can fork it off from here and leave it that way. If anyway ends up needing to use it like I did, they can find my fork and built it themselves or download pre-build binaries I can put on GitHub. I don't think there is any NuGet support in VS2008 so that is probably as good as it gets but it would work and it would be fine by me. I probably won't be needing this library myself in 12 months anyway, but I am sure someone else will find it useful :)

## Purpose
Port RestSharp to Windows Compact Framework

## Checklist


alexeyzimarev:
I really appreciate the effort but the issue as I see it is: we want to get to one project, with one codebase, with as least ifdefs as possible. There will be not many people using the Compact Framework, I'm afraid. To me the best solution for this would be to make a fork and merge this PR to that fork, may be make a package, but not merge it to RestSharp codebase.

Surely I would love to have a discussion around this.

alexeyzimarev:
I believe the best choice for Android is also use netstandard...

kendallb:
Yep, sounds fine with me. I will simply make a new fork and make it Compact Framework specific for anyone who wants to use it and build a new readme.md file that shows where to get Json.net that works on CF also. In a year I won't care about this either I hope :)

kendallb:
Ok for anyone else who stumbles across this and wants my port, you can get it here. This is now forked and will never end up in the main code tree, but for someone else who needs it I will do basic maintenance on it until I no longer need it myself :)

https://github.com/kendallb/RestSharp.CompactFramework

PR #1006, opened 2017-09-13
Removing obsolete stuff and moving tests to NUnit3

alexeyzimarev:
- Removing obsolete platforms:
   - .NET 3.5-4.5.1 - the earliest is 4.5.2
   - Silverlight
   - Windows Phone (both of them)
   - MonoTouch
- Updated NUnit to v3
- Fixed tests and build script to use NUnit3
- Fixed async tests to actually use async/await


RestSharp.Tests/JsonTests.cs
```diff
@@ -1060,7 +1062,7 @@ public string CreateDynamicJsonStringDictionary()
 
         private static T GetPayLoad<T>(string fileName)
         {
-            string doc = File.ReadAllText(Path.Combine("SampleData", fileName));
+            string doc = File.ReadAllText(Path.Combine(currentPath, "SampleData", fileName));
```
mjwsteenbergen: Why is this change neccesary?

alexeyzimarev: NUnit3 console runner does not guarantee to execute in the test library location. As a result, none of the test files are found by tests. You can try yourself or resolve it differently, I find this way the most efficient.

mjwsteenbergen: 👍

mjwsteenbergen:
No, it won't ;)

alexeyzimarev:
Thanks! Bold move.

PR #1008, opened 2017-09-19
update index.html

fredyfx:
Adding StackOverflow In Spanish link.

## Purpose
This pull request is a:

## Checklist

### DIFF
--- a/index.html
+++ b/index.html
@@ -34,7 +34,7 @@ title: RestSharp - Simple REST and HTTP Client for .NET
       <section>
 
         <h3>NEED HELP with RestSharp?</h3>
-        <p><strong>Post your question on <a href="http://stackoverflow.com/questions/tagged/restsharp">StackOverflow</a></strong>. If you think you've found a bug, create an issue on GitHub.</p>
+        <p><strong>Post your question on <a href="http://stackoverflow.com/questions/tagged/restsharp">StackOverflow</a></strong>. If you think you've found a bug, create an issue on GitHub. For Spanish support go to <a href="http://es.stackoverflow.com/questions/tagged/restsharp">StackOverflow en Español</a> </p>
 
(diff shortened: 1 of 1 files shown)


mjwsteenbergen:
Can you give a reason why we should add a spanish stackoverflow link? None of the maintainers are spanish and the new link you provide tell us that there are only 3 questions tagged with the restsharp tag of which one is not answered. For people seeking help, it would be beneficial not to send them to the spanish version of stackoverflow, but to the english version.

fredyfx:
**Can you give a reason why we should add a Spanish Stackoverflow link?**
Because, speaking for Hispanic developers, we want to increase the amount of developers who knows, uses and relies RESTful services on RESTSharp.

**None of the maintainers are spanish.**
I am able to be one. I accept the challenge.

**and the new link you provide tell us that there are only 3 questions tagged with the restsharp tag of which one is not answered.**
We are starting, this is the beginning. That one question which is not response is because 2 reasons: a) few people uses "mercado pago API". b) few users knows and uses RESTSharp. _Would you like to change this situation?_

**For people seeking help, it would be beneficial not to send them to the Spanish version of Stackoverflow, but to the English version.**
It is very comfortable to have people supporting you in your mother language. This applies for the Russian site and also the Portuguese site of StackOverflow.

What do you think now?

Thanks for reading,

Fredy.

mjwsteenbergen:
> It is very comfortable to have people supporting you in your mother language. 
I would like to thank you for helping those people, but as a non native speaker, I rather use the English version than my native language to search, as the knowledge base is way larger and usually of higher quality. Furthermore we would be splitting our own knowledge base and that would add more work than we are able to handle. This is why we will not merge this pull request. I do want to sincerely thank you for the interest in this project and wish you well,

Martijn

PR #1034, opened 2017-11-10
query string parameters are encoded correctly using the selected encoding instead of default encoding

masadwin:
No matter what encoding I selected, the Url query string parameters were
always encoded using the default UTFencoding.
Reason: i needed to call a GET method that required me to send my query
string encoded in ISO-8859-1 encoding. I tried many options but 'ø'
would always become '%C3%B8' even though I had set the encoding property
to ISO-8859-1 which should get me '%F8'
This change uses the currently selected encoding in the RestClient.cs
and uses it to encode parameters and names instead of using UrlEncode
which (although works for most cases) might not work for everybody.

### DIFF
--- a/RestSharp.Tests/StringExtensionsTests.cs
+++ b/RestSharp.Tests/StringExtensionsTests.cs
@@ -37,8 +37,8 @@ namespace RestSharp.Tests
         public void UrlEncodeTest()
         {
             const string parameter = "ø";
-            Assert.AreEqual("%F8", parameter.UrlEncode(Encoding.GetEncoding("ISO-8859-1")));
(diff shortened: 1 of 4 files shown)


masadwin:
moved to here from
https://github.com/restsharp/RestSharp/pull/843

i kind of messed up that repository and deleted it. later realizing that the synced changes were automatically added to commit without me needing to do anything. So this is a new one with same changes and a test case

masadwin:
okay it's finally in working condition. I feel like I should explain so many glorious commits. After the change I realized I need VS 2017 (currently using 2015) to open this project. So I did the changes on notepad++

And was using the build server to check if it is working fine

masadwin:
added another test to check impact on BuildUri

Issue #1035, opened 2017-11-13
BaseUrl is wrong when Resource is absolute (with protocol)


datvm:
In this issue: https://github.com/restsharp/RestSharp/issues/619 you are asking a Use case for `BaseUrl` should not null. I actually have a problem with IdentityServer (actually just a normal OAuth2 server):

`this.RestClient = new RestClient(this.Config.Authority);`

For example, the Url should be `https://www.example1.com/`

Then I perform a Discovery:

`var request = new RestRequest(".well-known/openid-configuration", Method.GET);`

And I can get the Endpoints. However, the response will be:

`"authorization_endpoint": "https://www.example1.com/connect/authorize",`

Now, I cannot use that RestClient, because:

`var request = new RestRequest(this.AuthorizeEndpoint, Method.GET);`

will returns this Uri:

`https://www.example1.com/https://www.example1.com/connect/authorize?`

The desirable should be, when a RestRequest resource is absolute, it should not be appended at all. I tried with Uri and this one is correct, why is it not the case for RestRequest?

```
            var a = new Uri("https://www.example1.com/");
            var b = new Uri(a, "https://www.example1.com/connect/authorize");
            var c = b.AbsoluteUri; // c is "https://www.example1.com/connect/authorize"
```

alexeyzimarev:
@datvm I merged the PR and the 106.1 package contains your fix. Could you please test if the issue is resolved and close the issue?

datvm:
Hi I have checked the 106.1. It's fixed. Thank you very much!

PR #1036, opened 2017-11-13
Fixed Uri builder when the Resource is an absolute Uri (with protocol, etc...).

datvm:
Details in the test case and this issue: https://github.com/restsharp/RestSharp/issues/1035

```
            RestRequest request = new RestRequest("https://www.example1.com/connect/authorize");
            
            RestClient client = new RestClient("https://www.example1.com/");
            Uri expected = new Uri("https://www.example1.com/connect/authorize");
```

### DIFF
--- a/RestSharp.Tests/UrlBuilderTests.cs
+++ b/RestSharp.Tests/UrlBuilderTests.cs
@@ -298,5 +298,17 @@ namespace RestSharp.Tests
             Assert.AreEqual(expected, output);
         }
 
+        [Test]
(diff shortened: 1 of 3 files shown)


datvm:
Sorry in the commit I forgot about the Project settings. Actually I don't know why the Debug profile has the Optimize Code on, so I could not debug the code. I turned it off to debug, and forgot to get it back on. Please turn it on again if you accept this pull request and the option is needed.

alexeyzimarev:
Optimise should be off for the Debug profile so this is ok.

Issue #1018, opened 2017-10-22
Make serialisation configurable


alexeyzimarev:
Currently, XML and JSON serialisation configuration is hard coded to existing implementations.

This needs to be changed in order to allow people using any kind of serialisers they want, by implementing serialisation interfaces and doing some minimal configuration.

This makes #994 obsolete.

alexeyzimarev:
Could also have something `UseSerializer<T>(ContentType.Json)` to solve #851

alexeyzimarev:
Done by now.

Issue #1033, opened 2017-11-10
Multidimensional array added to the body is converted to a 1d array


screig:
## Expected Behavior
I have a simple struct which contains a 2D array double[,]. I add this struct to the body of the POST request. 
`request.AddBody(myProblem);`
Looking in to the request object with Visual Studio (before it is sent) I can see that it has been converted to a 1D array in JSON.

If I convert the struct to JSON separately , like so
`string json = JsonConvert.SerializeObject(myProblem);`

I get this, which is correct. note the square brackets.

[...]

PR #1073, opened 2018-01-20
JsonSerializer to use Newtonsoft.Json instead of SimpleJson

Psypher9:
Since the documentation at the top of the JsonSerializer class defers to Newtonsoft.Json's attributes and most of the class is not actually implemented, simply use the Newtonsoft serializer directly.

## Description

An issue I and my team had been running into was with the JsonSerializer not recognizing the SerializeAs attribute (as is noted in the summary documentation above the JsonSerializer class) but also not recognizing the JsonProperty attribute from Newtonsoft and I could not find any references to Newtonsoft in the project dependencies. Adding a reference to Newtonsoft locally then testing our own project proved successful when we used the JsonConvert class to serialize our objects instead of SimpleJson. 

I had also noted that in issue #1033 the suggestion was posited to use Json.NET to serialize instead the JsonSerializer provided at the moment.

## Purpose
This pull request is a:

- [x ] Bugfix (non-breaking change which fixes an issue)

## Checklist

- [x ] I have added necessary documentation (if appropriate)

### DIFF
--- a/RestSharp/RestSharp.csproj
+++ b/RestSharp/RestSharp.csproj
@@ -18,6 +18,7 @@
     <PackageReference Include="GitVersionTask" Version="4.0.0-beta*">
       <PrivateAssets>All</PrivateAssets>
     </PackageReference>
+    <PackageReference Include="Newtonsoft.Json" Version="10.0.3" />
     <PackageReference Include="System.ValueTuple" Version="4.4.0" />
   </ItemGroup>
   <ItemGroup>
--- a/RestSharp/Serializers/JsonSerializer.cs
+++ b/RestSharp/Serializers/JsonSerializer.cs
@@ -1,9 +1,12 @@
-﻿
+﻿using Newtonsoft.Json;
+
 namespace RestSharp.Serializers
 {
     /// <summary>
     /// Default JSON serializer for request bodies
     /// Doesn't currently use the SerializeAs attribute, defers to Newtonsoft's attributes
+    /// To Use Newtonsoft's attributes in place of the SerializeAs attribute use
+    /// [JsonProperty(PropertyName = "<name>")]
     /// </summary>
     public class JsonSerializer : ISerializer
     {
@@ -22,7 +25,7 @@ namespace RestSharp.Serializers
         /// <returns>JSON as String</returns>
         public string Serialize(object obj)
         {
-            return SimpleJson.SimpleJson.SerializeObject(obj);
+            return JsonConvert.SerializeObject(obj);
         }
 
         /// <summary>


Psypher9:
After submitting this pull request and thinking about for a few days, I may have an idea to extend Newtonsoft's attributes so that RestSharp can fully support the [SerializeAs] attribute.

alexeyzimarev:
Thanks for the PR but for most users SimpleJson works. I realise the issues with it and I know that for many people using Json.Net would be better. Therefore, I have plans to improve the serialisation abstraction. When this will be done, an additional library can be made to add Json.Net using some simple initialisation for the `RestClient`

alexeyzimarev:
This is the issue to fix #1018

alexeyzimarev:
It is possible to change the default deserialiser in the client by using this method:

https://github.com/restsharp/RestSharp/blob/develop/RestSharp/RestClient.cs#L217

alexeyzimarev:
@Shikada if you are so tired why don't you submit a PR to fix it in a proper way?

Shikada:
I was under the impression that this was already being worked on. In google groups I once proposed a solution to a bug I found, and you said:

> I'd rather just change serialisation to use Json.Net, it is not hard to do, there is a package for that and I plan to make one in the core project too.

So yeah, I don't want to just throw negative comments your way and expect you to do the work I need. If you confirm that you want this done, I will try to find the time to make and submit a solution. I would just like to avoid working on this if you're already in the process of making this change.

alexeyzimarev:
I wrote "I have plans". I have not figured out how to do this change that it is non-breaking. If you have any idea about this - I am happy to discuss.

Psypher9:
@alexeyzimarev are you still wanting to use Json.NET or make a custom serializer?

alexeyzimarev:
@Psypher9 yes, I want to get back to Json.NET but I need to change the way how serialization is configured so people can use something else. There are plenty of fast JSON serialization libraries now.

Psypher9:
@alexeyzimarev ah I see. Got it!

Psypher9:
@alexeyzimarev I would just like to revisit this and say that I agree with you that there are big benefits for the developer to have control over which serializer they use; however, what I was trying to accomplish with this pull request was to simply replace the default serializer because the SimpleJson library wasn't supporting the Newtonsoft attributes for Serialization (`[JsonProperty(PropertyName = "propName")]`).  The `RestClient.JsonSerializer` is still configurable by the developer as it stands now. My team and I have been able to replace it by creating a new class that implements `ISerializer` like this:

here's the class that implements `ISerializer`
```csharp
    public class NewtonsoftRestSharpSerializer : ISerializer
```
Then it is used in the request like this
```csharp
            RestRequest request = new RestRequest(path, method);
            request.JsonSerializer = new NewtonsoftRestSharpSerializer();
            request.RequestFormat = DataFormat.Json;
            return request;
```
Is this different than what you had in mind? Something like a constructor overload for `RestRequest`, maybe? Because I'd be willing to help out with that and perhaps work on the `SerializeAs` attribute so that it can work as intended with JSON serialization.

alexeyzimarev:
Yeah, the issue with this is that you do too much work for each request. It should be possible to define the `RestClient` that uses certain serializer for all requests.

alexeyzimarev:
I also think that avoiding taking a dependency on Json.Net makes no sense, considering the number of issues caused by using SimpleJSON, which is heavily outdated and is not being worked on. But I realize that Json.Net is also the slowest serialzier out there despite being the most popular one, so it must be possible for people to easily use their own.

PR #1086, opened 2018-02-23
Using the exact name provided in DeserializeAs attribute, if any, on …

Andrius-Dobrotinas:
## Description

…Deserialization

When using DeserializeAs attribute on a class property, I specified a name that the attribute/node is supposed to have (which included a dash between words). Deserialization didn't work. I removed a DeserializeAs attribute altogether, and deserialization worked. This means that there was a bug in the use of the DeserializeAs.Name property. The way I see it, if the name is explicitly specified in DeserializeAs attribute, the deserializer should use that value.

## Purpose
This pull request is a:

Seems to be non-breaking change as none of the pre-existing tests failed after the modification.

## Checklist

### DIFF
--- a/RestSharp.Tests/NamespacedXmlTests.cs
+++ b/RestSharp.Tests/NamespacedXmlTests.cs
@@ -22,6 +22,7 @@ using System.Xml.Linq;
 using NUnit.Framework;
 using RestSharp.Deserializers;
 using RestSharp.Tests.SampleClasses;
+using RestSharp.Tests.SampleClasses.DeserializeAsTest;
 using RestSharp.Tests.SampleClasses.Lastfm;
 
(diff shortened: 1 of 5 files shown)


Andrius-Dobrotinas:
It seems like the build has failed due to an issue unrelated to my fix: RestSharp.IntegrationTests.AsyncTests.Can_Cancel_GET_TaskAsync. What do I do about that?

Andrius-Dobrotinas:
@alexeyzimarev 
It seems like the build has failed due to an issue unrelated to my fix: RestSharp.IntegrationTests.AsyncTests.Can_Cancel_GET_TaskAsync. What do I do about that?

alexeyzimarev:
I need to look at the tests.

Issue #1183, opened 2018-10-04
HttpMethod is not overridden in `RestClient.Sync.Execute(IRestRequest request, Method httpMethod)`


Psypher9:
## Expected Behavior
I expected this method to override the Request's HttpMethod as documented, but it was not.

## Actual Behavior
The HttpMethod is not overridden.

## Steps to Reproduce the Problem

  1. The code is self evident.

alexeyzimarev:
Thanks for the PR! It is out in 106.5

Issue #1182, opened 2018-10-03
Does Not work with Windows 10 Version 10.0.17763.1 


czupet:
## Actual Behavior
Error: 
Operation is not supported on this platform.
in response variable.

## Steps to Reproduce the Problem

  1.
 var request = new RestRequest("getserver/data1")
                {
                    OnBeforeDeserialization = resp => { resp.ContentType = "application/json"; }
                };

                IRestResponse<string> response = await _client.ExecuteGetTaskAsync<string>(request, _cancellationTokenSource.Token);
  2.
  3.

## Specifications

  - Version:
106.4.2
  - Platform:
Windows 10 Version 10.0.17763.1 
UWP Program.
  - Subsystem:
64bit

## StackTrace
{System.PlatformNotSupportedException: Operation is not supported on this platform.
   at System.Net.SystemWebProxy.GetProxy(Uri destination)
   at System.Net.ServicePointManager.ProxyAddressIfNecessary(Uri& address, IWebProxy proxy)
   at System.Net.ServicePointManager.FindServicePoint(Uri address, IWebProxy proxy)
   at System.Net.HttpWebRequest.get_ServicePoint()
   at RestSharp.Http.ConfigureWebRequest(String method, Uri url)
   at RestSharp.Http.ConfigureAsyncWebRequest(String method, Uri url)
   at RestSharp.Http.GetStyleMethodInternalAsync(String method, Action`1 callback)}

<details>

V106.4.2 for UWP Does not work with previous version of Windows 10 either, 
RestSharp V106.3.1 does work with previous version of Windows 10, but does not work with Version 10.0.17763.1 

</details>

Enngage:
I'm getting the same exception :(

czupet:
And of course compiled application will not work also.  So if you had an application that was working and updated Windows 10 to V 10.0.17763.1, your application will not connect to WebAPI.

Psypher9:
Possibly a duplicate of #1179?

czupet:
It's possible, but this is an UWP application.

alexeyzimarev:
If you all cooperate, maybe you can fix this? I have no Windows.

czupet:
Unfortunately like most of the people here I have not time to spare, I have a project going into production this coming Monday, and I am just hopping nobody has installed the new version of Windows.
[...]

Issue #1061, opened 2018-01-01
"Operation not supported on this platform" on Core 2, stable release 106.2.0


erlando:
## Expected Behavior
Request should be performed

## Actual Behavior
Request is not performed. Error is encountered

## Steps to Reproduce the Problem
    var client = new RestClient("http://example.com");
    var request = new RestRequest("/", Method.GET);
    var response = client.Execute(request);
    
    // response.IsSuccessful == false, response.ErrorMessage == "Operation not supported on this platform"

## Specifications

  - Version: 106.2.0
  - Platform: .Net Core 2.0

**No proxy is defined**

## StackTrace
<details>
System.PlatformNotSupportedException: Operation is not supported on this platform.
   at System.Net.SystemWebProxy.GetProxy(Uri destination)
   at System.Net.ServicePointManager.ProxyAddressIfNecessary(Uri& address, IWebProxy proxy)
   at System.Net.ServicePointManager.FindServicePoint(Uri address, IWebProxy proxy)
   at System.Net.HttpWebRequest.get_ServicePoint()
   at RestSharp.Http.ConfigureWebRequest(String method, Uri url)
   at RestSharp.Http.GetStyleMethodInternal(String method)
   at RestSharp.Http.AsGet(String httpMethod)
   at RestSharp.RestClient.DoExecuteAsGet(IHttp http, String method)
   at RestSharp.RestClient.Execute(IRestRequest request, String httpMethod, Func`3 getResponse)</details>

erlando:
#1010

alexeyzimarev:
I need a stack trace.

alexeyzimarev:
Sorry, found it.

Raylehnhoff:
I had this same issue and found this thread. Here's the stack trace from the response.ErrorException property.

There was no proxy.

Using the prior alpha version referenced in #1010 fixes this. Using the latest version does not.

````
{System.PlatformNotSupportedException: Operation is not supported on this platform.
   at System.Net.SystemWebProxy.GetProxy(Uri destination)
   at System.Net.ServicePointManager.ProxyAddressIfNecessary(Uri& address, IWebProxy proxy)
   at System.Net.ServicePointManager.FindServicePoint(Uri address, IWebProxy proxy)
   at System.Net.HttpWebRequest.get_ServicePoint()
   at RestSharp.Http.ConfigureWebRequest(String method, Uri url)
   at RestSharp.Http.GetStyleMethodInternal(String method)
[...]

PR #1184, opened 2018-10-04
Some bugfixes and interface segregation

Psypher9:
## Description

Issues this pull request addresses:
- #1183 
- #1179 
- #1182 
- #1061

## Purpose
This pull request is intended to fix the issues mentioned above as well as segregating the `ISerializer` interface into two interfaces: `ISerializer` and `IXmlSerializer`. The reason for this segregation is due to the need to stub the XML specific properties on a JSON serializer.

- [ x ] Bugfix (non-breaking change which fixes an issue)

## Checklist

- [ x ] I have added tests that prove my fix is effective or that my feature works
- [ x ] I have added necessary documentation (if appropriate)


Psypher9:
I just realized that my fix for the HttpMethod override in the RestClient had a merge error when I merged my branches. It's not broken, but I'm going to fix that.

RestSharp/RestClient.Sync.cs
```diff
@@ -40,6 +40,7 @@ public virtual IRestResponse Execute(IRestRequest request, Method httpMethod)
         {
             if (request == null)
                 throw new ArgumentNullException(nameof(request));
+            request.Method = httpMethod;
```
alexeyzimarev: What's the point of this change?

alexeyzimarev:
I love the fact that you took care of fixing the proxy issue. At the same time, doing so many changes in one PR makes it hard to review, especially when it includes formatting changes that are distracting from the functional changes. I will spend more time reviewing it tomorrow.

RestSharp/RestClient.Sync.cs
Psypher9: This is actually something that I need to fix, before moving forward. When I added that line there was not a line to override the HTTP Method on the request as this method is documented to do on [Line 37](https://github.com/restsharp/RestSharp/blob/29b7a38a3ee2478311f2a7d7c954d56d4a37f914/RestSharp/RestClient.Sync.cs#L37). I was reading through the `RestClient` and studying its structure when I happened to notice that the Method was not being overridden, the `httpMethod` parameter was being passed in but never used.

It appears that there was a merge problem between my `master` and `develop` branches, making it look like there are two lines, so that's what I'm going to fix.

Psypher9: Yeah it actually looks like the commit right before this line that I added had done the same thing and I just missed it. So I'll revert that change and push back up.

Psypher9:
Okay @alexeyzimarev, I fixed that unnecessary change in `RestClient.Sync`

RestSharp/RestRequest.cs
```diff
@@ -74,10 +74,16 @@ public RestRequest(string resource) : this(resource, Method.GET)
         /// </summary>
         /// <param name="resource">Resource to use for this request</param>
         /// <param name="method">Method to use for this request</param>
-        public RestRequest(string resource, Method method) : this()
+        public RestRequest(string resource, Method method) : this(resource, method, DataFormat.Xml)
```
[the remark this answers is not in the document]
Psypher9: Do you still want to discuss this change?

jeanpoelie:
Error still exists for me in the newest version

106.5.2
netcore 2.1
Windows 10

at System.Net.SystemWebProxy.GetProxy(Uri destination) at System.Net.ServicePointManager.ProxyAddressIfNecessary(Uri& address, IWebProxy proxy) at System.Net.ServicePointManager.FindServicePoint(Uri address, IWebProxy proxy) at System.Net.HttpWebRequest.get_ServicePoint() at RestSharp.Http.ConfigureWebRequest(String method, Uri url) at RestSharp.Http.PostPutInternal(String method) at RestSharp.Http.AsPost(String httpMethod) at RestSharp.RestClient.DoExecuteAsPost(IHttp http, String method) at RestSharp.RestClient.Execute(IRestRequest request, String httpMethod, Func3 getResponse)`

PR #1202, opened 2018-10-18
Fix #1188 scraping get parameters already added

### LINKED ISSUE #1188: OAuth1 signature_invalid when adding query parameters
## Expected Behavior
For requests to authenticate.

## Actual Behavior
OAuth failure message: signature_invalid
I don't think it's related to #1166

## Steps to Reproduce the Problem

  0. `var _client = new RestClient(EtsyApiEndpoint) { Authenticator = OAuth1Authenticator.ForProtectedResource(appKey, appSecret, accessKey, accessSecret) };`
  1. `var req = new RestRequest("/shops/{shopId}/listings/active");`
  2. `req.AddParameter("includes", "Attributes,Inventory");`
  3. `var result = _client.Execute(req);`

## Specifications

  - Version: 106.5.0
  - Platform: Windows
  - Subsystem: Console App?

## Details

I think the problem is at [OAuth1Authenticator.AddOAuthData](https://github.com/restsharp/RestSharp/blob/master/RestSharp/Authenticators/OAuth1Authenticator.cs#L217) (adding get or post params) and [OAuthWorkflow.OAuthWebQueryInfo](https://github.com/restsharp/RestSharp/blob/master/RestSharp/Authenticators/OAuth/OAuthWorkflow.cs#L218) (adding querystring params), so request parameters that became part of the querystring are being doubled.  Stepping over the second part during debugging seems to work.

zaus:
## Description

Fixes #1188 

## Purpose
This pull request is a:

## Checklist

I can't see how to verify that the signature was created correctly _via unit test_, but I eyeballed it against a verifier tool like http://lti.tools/oauth/


artiomchi:
Hmm.. My concern with this is that it won't allow you to add multiple query string parameters with the same name. While this might not be an issue for everyone, it might be an issue for some :(

PR #1208, opened 2018-10-23
Better use of AdvancedResponseWriter

yurikus:
## Description

1. Moved ProcessResponseStream() call to the bottom of ExtractResponseData() to make HttpResponse members available to AdvancedResponseWriter.
2. Check for AdvancedResponseWriter first and call only that or fallback to ResponseWriter.

## Purpose
This pull request is a:

- [X] Improvement

### DIFF
--- a/RestSharp/Http.cs
+++ b/RestSharp/Http.cs
@@ -390,10 +390,6 @@ namespace RestSharp
                 response.ContentType = webResponse.ContentType;
                 response.ContentLength = webResponse.ContentLength;
 
-                var webResponseStream = webResponse.GetResponseStream();
-
-                ProcessResponseStream(webResponseStream, response);
-
                 response.StatusCode = webResponse.StatusCode;
                 response.StatusDescription = webResponse.StatusDescription;
                 response.ResponseUri = webResponse.ResponseUri;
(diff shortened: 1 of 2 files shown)


RestSharp/Http.cs
```diff
@@ -430,18 +426,24 @@ private void ExtractResponseData(HttpResponse response, HttpWebResponse webRespo
...
+
                 webResponse.Close();
             }
         }
 
         private void ProcessResponseStream(Stream webResponseStream, HttpResponse response)
         {
-            if (ResponseWriter == null)
-                response.RawBytes = webResponseStream.ReadAsBytes();
+            if (AdvancedResponseWriter != null)
```
alexeyzimarev: What is the purpose of this change? For this method, the previous code did exactly the same, didn't it?

yurikus: It did not to the same thing. When ResponseWriter is null and AdvancedResponseWriter is non-null, stream will be consumed by ReadAsBytes() call.

alexeyzimarev:
Thank you for the PR! :+1:  In order to merge this, I need to have a unit test that is failing on the existing version and goes green for the PR, so we know what value it brings.

PaulVrugt:
This fix would also help us a lot. Please hurry ;)

yurikus:
I added a test.

PaulVrugt:
so, when are we releasing this? We are anxiously awaiting this fix

PaulVrugt:
@alexeyzimarev can we create a release containing this fix?

PR #1213, opened 2018-10-29
Allow setting underlying HttpWebRequest's AllowWriteStreamBuffering to prevent OutOfMemoryExceptions

### LINKED ISSUE #889: Out of Memory issue when uploading large files
When we are uploading the large files with RestSharp, we randomly get some OutOfMemory issues. This happens because RestSharp is using HttpWebRequest behind the scene, and the default value for AllowWriteStreamBuffering is true. This mean that the whole file are loaded in memory when doing the query, in case there are some redirections to be able to replay it. But If we are uploading a 500 mb file, its an issue.

What we would need is to be able to set the AllowWriteStreamBuffering to false, but there is no way right now to alter the AllowWriteStreamBuffering in the ConfigureWebRequest.

Something like this would be sufficient for our need:

`var request = new RestRequest(Method.POST);
request.AllowWriteStreamBuffering = false;
request.Files.Add(new FileParameter // ...`

yvesmh:
## Description

Fixes #889

## Purpose
This pull request is a:

## Checklist

### DIFF
--- a/RestSharp/Http.Sync.cs
+++ b/RestSharp/Http.Sync.cs
@@ -199,6 +199,7 @@ namespace RestSharp
             webRequest.UseDefaultCredentials = UseDefaultCredentials;
 
             webRequest.PreAuthenticate = PreAuthenticate;
+            webRequest.AllowWriteStreamBuffering = AllowWriteStreamBuffering;
             webRequest.Pipelined = Pipelined;
             webRequest.UnsafeAuthenticatedConnectionSharing = UnsafeAuthenticatedConnectionSharing;
 #if NETSTANDARD2_0
--- a/RestSharp/Http.cs
+++ b/RestSharp/Http.cs
@@ -211,6 +211,12 @@ namespace RestSharp
         public IList<DecompressionMethods> AllowedDecompressionMethods { get; set; }
 
         /// <summary>
+        ///     Sets the underlying HttpWebRequest's AllowWriteStreamBuffering flag so large files
+        ///     don't need to be loaded into memory, causing OutOfMemoryExceptions
+        /// </summary>
+        public bool AllowWriteStreamBuffering { get; set; }
+
+        /// <summary>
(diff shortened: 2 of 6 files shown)


alexeyzimarev:
We have the `ConfigureWebRequest` method that allows using any additional configuration for specific cases, do we need to keep adding more stuff to RestSharp for handling edge cases?

yvesmh:
Hi @alexeyzimarev, the method `ConfigureWebRequest` is protected and I couldn't find anywhere in the code where RestSharp exposes the `HttpWebRequest` that `ConfigureWebRequest` returns, which is why I added the property to the other classes. Is there a cleaner solution in your mind? I'm at work right now but I can work on this later today.

alexeyzimarev:
It's on `IRestClient`. `client.ConfigureWebRequest(x => x.AllowWriteStreamBuffering = true);`

alexeyzimarev:
If we set this property to true by default instead of configuring it, what is the downside?

alexeyzimarev:
The docs say the default is true...

alexeyzimarev:
Ah, I see now. It should be false.

yvesmh:
My bad, I was looking at Http's `ConfigureWebRequest` instead of `IRestClient`. I see that it is possible to configure the Web Request now.

Since issue #889 is labeled "good first issue"  and "help wanted" I thought the way to fix it was by exposing the flag.

I agree that for handling edge cases such as this one, it's best to just use `ConfigureWebRequest`. Would you like me to close this PR? The issue could also be closed with explanation in case anyone lands on it while troubleshooting similar issues.

alexeyzimarev:
I appreciate the PR. I added the configuration method since the number of small things that can be set on `WebRequest` is huge and repeating all these properties in the `IRestRequest` and `IRequestClient` is very troublesome. I was unable to find all cases that can be solved using this delegate. Let's close this one and indeed people that experience this issue could refer to it. I will also close the issue with an explanation.

PR #1214, opened 2018-10-29
An array of objects is also valid Json and should be handled by JsonSerializer. 

Bumler:
## Description
This is an expansion on https://github.com/restsharp/RestSharp/pull/1201 which allowed JsonSerializer to handle already serialized strings.

[ { "key", "value"} ] Is also valid json and could be sent as a string. The Serialize method should handle all valid json formats.

I chose to move IsSerializedString to its own method because it was beginning to create readability issues in the Serialize method.

Please let me know if anything needs to be cleaned up in my changes!

## Purpose
This pull request is a:

## Checklist


Bumler:
Will reopen with updated PR

PR #1230, opened 2018-11-28
Allow null resource in RestRequest constructor

oleveau:
## Description
Trying to create a RestRequest with a null resource throws an NullReferenceException

## Purpose
This pull request is a:

- [X ] Bugfix (non-breaking change which fixes an issue)

## Checklist

- [ X] I have added tests that prove my fix is effective or that my feature works


alexeyzimarev:
I fixed it a bit differently, thanks for pointing out.

PR #1282, opened 2019-02-22
Make SourceLink a development dependency only

jnyrup:
## Description

106.6.8 exposes a unnecessary dependency on `Microsoft.SourceLink.GitHub`.

From https://github.com/dotnet/sourcelink#githubcom-and-github-enterprise
> Source Link package is a development dependency, which means it is only used during build. It is therefore recommended to set PrivateAssets to all on the package reference. This prevents consuming projects of your nuget package from attempting to install Source Link.

## Purpose
This pull request is a:

## Checklist


alexeyzimarev:
Cool, thanks. I didn't think about that, thank you for fixing.

PR #1299, opened 2019-04-03
Update MiscExtensions.cs

blacksnake-rus:
Fix method AsString for encoding = null

## Description

Fix method AsString for encoding = null

## Purpose
This pull request is a:

### DIFF
--- a/RestSharp/Extensions/MiscExtensions.cs
+++ b/RestSharp/Extensions/MiscExtensions.cs
@@ -86,14 +86,17 @@ namespace RestSharp.Extensions
         /// <returns>The byte as a string.</returns>
         public static string AsString(this byte[] buffer, string encoding)
         {
-            Encoding enc;
+            Encoding enc = Encoding.UTF8;
             try
             {
-                enc = Encoding.GetEncoding(encoding);
+                if (!string.IsNullOrEmpty(encoding))
+                {
+                    enc = Encoding.GetEncoding(encoding);
+                }
             }
             catch (Exception)
             {
-                enc = Encoding.UTF8;
+
             }
 
             return AsString(buffer, enc);


IanBuswell-etech:
Be nice to get this in as the exception being thrown in the background was a large annoyance when trying to debug code that uses RestSharp.

alexeyzimarev:
@IanBuswell-etech I am not sure what you mean. You can easily turn off debugging external code in any .NET IDE.

alexeyzimarev:
But I agree that it is better to prevent the exception. However, this code will _still_ throw if the encoding contains something that doesn't resolve to a valid encoding. I don't see any difference, could you please explain?

blacksnake-rus:
> But I agree that it is better to prevent the exception. However, this code will _still_ throw if the encoding contains something that doesn't resolve to a valid encoding. I don't see any difference, could you please explain?

In most cases, the encoding is either correct or null. Cases where the encoding is not recognized are very rare. I do not want to spend CPU time to the unfolding of the full stacktrace at constant drops in the cases where the encoding is null.

FGS-BMarkham:
Is this a possible fix for https://github.com/restsharp/RestSharp/issues/1321 ?

blacksnake-rus:
> Is this a possible fix for #1321 ?

Quite possible!
Can you check in your application?

Metritutus:
To avoid it throwing exceptions for invalid encodings, and the need for an empty catch block at all, you could potentially change it to the following:

```csharp
            if(!string.IsNullOrWhiteSpace(encoding))
            {
                enc = Encoding.GetEncodings()
                                .FirstOrDefault(e => string.Equals(e.Name, encoding, StringComparison.OrdinalIgnoreCase))
                                ?.GetEncoding() ?? Encoding.UTF8;
            }
```

Krimsar:
As far as I can tell from quickly looking at the code, the wrong property is actually passed to this method. I think there's a misunderstanding of what the ContentEncoding property of a HttpWebResponse actually means. It is used for additional encoding of the body content, like gzipping it. (see https://blogs.msdn.microsoft.com/wndp/2006/08/21/content-encoding-content-type/) It will normally not be a value that can be passed to Encoding.GetEncoding. In fact it is null in many cases which results in an exception being thrown on every request. What you are looking for is the ContentType property (e.g.: "application/json; charset=utf-8") or more specifically the CharacterSet property, which gives you "utf-8".

sergei66666:
I agree with @Krimsar. As [link](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Encoding) says ContentEncoding only for compression.
So, there are should be next steps:
1) Replace all `RawBytes.AsString(ContentEncoding)` with `RawBytes.AsString(ExtractCharsetFromContentType(ContentType), RestClientInstance.Encoding)`.
But I'am not sure how to pass correctly Encoding property from instance of RestClient.
2) Implement method `ExtractCharsetFromContentType` which will extract charset from ContentType or will return null.
3) Changes to `RawBytes.AsString`:
if charset is null or empty use RestClientInstance.Encoding,
if exception when search encoder use RestClientInstance.Encoding.

dme-development:
When will this be released in a new nuget package?

alexeyzimarev:
Currently, the release is blocked because tests are hanging in the pipeline. I spent a couple of days lately trying to fix this issue but it is still work in progress.

alexeyzimarev:
Merged

Issue #1300, opened 2019-04-05
NullReferenceException for body parameters that don't have a name


stodolos:
## Expected Behavior
I would not have to add a name, or the error message would be clearer that a parameter name was missing.

## Actual Behavior
A NullReferenceException is thrown.

## Steps to Reproduce the Problem

  1. Add a request body parameter without a name

This must be the issue in RestClient.ConfigureHttp (it's assuming the name is there):

```
var parameterExists =
                    request.Parameters.Any(p =>
                        p.Name.Equals(defaultParameter.Name, StringComparison.InvariantCultureIgnoreCase)
                        && p.Type == defaultParameter.Type);
```

## Specifications

  - Version: 106.6.9.0
  - Platform: .NET Framework
  - Subsystem:

## StackTrace
<details>
System.NullReferenceException: Object reference not set to an instance of an object.
   at RestSharp.RestClient.<>c__DisplayClass172_0.<ConfigureHttp>b__8(Parameter p)
   at System.Linq.Enumerable.Any[TSource](IEnumerable`1 source, Func`2 predicate)
   at RestSharp.RestClient.ConfigureHttp(IRestRequest request)
   at RestSharp.RestClient.ExecuteAsync(IRestRequest request, Action`2 callback, String httpMethod, Func`4 getWebRequest)
   at RestSharp.RestClient.ExecuteAsync(IRestRequest request, Action`2 callback, Method httpMethod)
</details>

alexeyzimarev:
@vbenedichuk any reason why you don't want to submit a PR?

alexeyzimarev:
Closed in the next release. Breaking change!

mitcht:
I was attempting to debug / fix a problem but found that your change introduced a bug.  See line https://github.com/restsharp/RestSharp/blob/a8a9f3473be30317fc449c1c47e3f3cd74e8bab4/RestSharp/RestRequest.cs#L335

You'll see that the line is 
AddParameter(new JsonParameter("", obj));

Not sure how you'd like to go about handling this, but i'll not submit a pull request for something changed so recently.  Thanks!

alexeyzimarev:
Seems like we are missing some tests. Indeed body parameters don't need a name. I think a better fix would only check if the name is not empty for parameters that aren't of type body.

mitcht:
I updated the pull request to reflect your idea.

mitcht:
[...]

PR #1302, opened 2019-04-06
Null parameters

vbenedichuk:
## Description

Fix for #1300 NullReferenceException for parameters without a name

## Purpose
This pull request is a:

## Checklist


RestSharp/RestClient.cs
```diff
@@ -524,6 +524,10 @@ private string EncodeParameter(Parameter parameter, Encoding encoding)
 
         internal IHttp ConfigureHttp(IRestRequest request)
         {
+            if (request.Parameters.Any(p => p.Name == null))
+            {
+                throw new ArgumentNullException("null", "Parameter name is missing. Please make sure that all the parameters names are not null.");
```
alexeyzimarev: The `ArgumentNullException` first parameter is the parameter name. Here we have dynamic parameters, I don't think it is a good idea to have this exception here. If we want to check it, it is better to throw this exception before the request is executed, in `AddXXXParameter` and there the `name` would make sense.

vbenedichuk: I've thought about it but the problem that there are several ways to add parameter.
`request.Parameters.Add(new Parameter(...))`
`request.AddParameter(name, value)`
`request.AddParameter(new Parameter(...))`
and so on.
Adding check to every method will make it's support too complex.
I thought about adding this exception into Parameter class but it might be necessary to create a Parameter with empty name and set it later, exception will break this case.
ConfigureHttp looks like the only place where it can be solved with single condition.

alexeyzimarev: The issue is that it is much better to fail early. The logical question would be - why can I _add_ the parameter, but it fails on executing the request? And this question is completely legit.

vbenedichuk: I'm totally agree with you but current parameters management does not allow that. Feel free to provide better PR.

alexeyzimarev: I think it does because the easiest way would be to check if the name is supplied when the new instance of the `Parameter` class is created.

vbenedichuk: Thank you.

PR #1306, opened 2019-04-20
Fix setting deserializer's RootElement from request's RootElement

mowcixo:
## Description

When you set request's RootElement property, that's not setting back in deserializer's RootElement.

## Purpose
This pull request is a:

## Checklist


mowcixo:
I'm aware that RootElement from request is currently deprecated and it's going to be deleted, but is not working right now.

By the way, is there any example that shows how to replace that workaround when that property is deleted?

Issue #1305, opened 2019-04-10
Adding Content-MD5 header with deferred body serialization


cdaven:
We have a custom `IAuthenticator` implementation that, among other things, adds a "Content-MD5" header with an MD5 hash of the request's body. This stopped working in version 106.6 (?), and this is why:

> To prepare for this change, we made quite a few changes in how serialization works in RestSharp. **Before, objects were serialized when added to the RestRequest by using one of the AddBody methods.** That made it impossible to assign a custom serializer on the client level, so it should have been done for each request. **In v106.6 body parameter is serialized just before executing the request.** Delaying the serialization allowed us to add the client-level serializer.

Our code looks something like this:

```csharp
class CustomAuthenticator : IAuthenticator
{
    public void Authenticate(IRestClient client, IRestRequest request)
    {
        var body = request.Parameters.FirstOrDefault(p => p.Type == ParameterType.RequestBody);
        request.AddHeader("Content-MD5", md5(body.Value as string));
    }
}
```

The problem in version 106.6 is that `body.Value` is no longer serialized to a JSON string, so we cannot compute a hash sum yet.

We _could_ serialize the body in the authenticator ourselves, and _hope_ that the actual body in the request will be serialized to an identical string, but that seems a bit fragile, and wasteful.

Is there a better solution to this?

alexeyzimarev:
I guess adding something similar to https://github.com/restsharp/RestSharp/blob/master/RestSharp/IRestRequest.cs#L375 would solve this issue?

cdaven:
I have modified our code to work with the above callback instead, reading the serialized `http.RequestBody` and setting the Content-MD5 header with `http.Headers.Add(new HttpHeader { Name = "Content-MD5", Value = md5(http.RequestBody) });`.

It works, but feels a bit like a "hack" to not use the intended authentication mechanism.

alexeyzimarev:
How would you solve the issue? I really want to keep client-level serializers and it means that we cannot serialize when adding parameters because at that point we have no reference to the client.

Shall we try moving the authentication call?

ewahner:
@alexeyzimarev not sure if this is the right place to ask this, but your solution does not seem to work with Files as the RequestBody and RequestBodyBytes is null in `OnBeforeRequest` .  Is there another way to accomplish this with Files?

alexeyzimarev:
[...]

PR #1312, opened 2019-05-09
Add callback/event RestRequest.OnBeforeRequest

cdaven:
## Description

The callback/event `OnAfterConfiguration` is executed after the HTTP object is configured, but before the HTTP request is made. This allows for some last minute header modifications, which in my case is necessary due to the delayed serialization in 106.6 and later.

I've tested using `OnAfterConfiguration` instead of setting `RestClient.Authenticator`, and it seems to do the trick.

Solves issue #1305 

## Purpose
This pull request is a:

## Checklist

### DIFF
--- a/RestSharp/IRestRequest.cs
+++ b/RestSharp/IRestRequest.cs
@@ -374,6 +374,8 @@ namespace RestSharp
 
         Action<IRestResponse> OnBeforeDeserialization { get; set; }
 
+        Action<IHttp> OnAfterConfiguration { get; set; }
+
         void IncreaseNumAttempts();
     }
 }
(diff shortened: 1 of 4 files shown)


alexeyzimarev:
Nice addition, thank you. One question - is `OnAfterConfiguration` a good enough name, maybe we can call it `OnBeforeRequest`?

cdaven:
> Nice addition, thank you. One question - is `OnAfterConfiguration` a good enough name, maybe we can call it `OnBeforeRequest`?

Yes, that seems like a better name for it.

alexeyzimarev:
I fixed the potential null reference exception, you were calling the delegate even if it wasn't supplied.

PR #1323, opened 2019-06-13
Allows ISO-8859-1 encoding for basic authentication

jairbubbles:
## Description

*I've been experimenting [authentication issues ](https://bitbucket.org/farmas/atlassian.net-sdk/wiki/Home) with Jira .NET Api (which uses RestSharp under the hood).  When using french character 'é' I couldn't authenticate. It turns out that the Jira server expects ISO-8859-1 encoding.*

This pull request allows to use ISO-8859-1 encoding for the BASIC header.  Default encoding remains UTF-8.

In this thread you'll find some more info: https://stackoverflow.com/questions/7242316/what-encoding-should-i-use-for-http-basic-authentication  

## Purpose
This pull request is a:

## Checklist

### DIFF
--- a/RestSharp/Authenticators/HttpBasicAuthenticator.cs
+++ b/RestSharp/Authenticators/HttpBasicAuthenticator.cs
@@ -22,15 +22,26 @@ using System.Text;
 
 namespace RestSharp.Authenticators
 {
+    /// <summary>
+    /// Allows "basic access authentication" for HTTP requests.
+    /// </summary>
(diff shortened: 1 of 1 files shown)


alexeyzimarev:
I am not sure if we want to have a special case for one specific encoding. What do we do with all other potential encodings? I'd suggest making it more generic.

alexeyzimarev:
I changed your code so you supply the encoding you want.

PR #1337, opened 2019-07-22
Fixed breaking change so object initialized Parameter classes do not …

mitcht:
Description
I've noticed that a recent change of parameter name validation breaks existing code.  I've added a new base class constructor that does not use the validation for the XmlParameter and JsonParameter inherited classes, and assigned the name in them respectively. 

Purpose
This pull request is a:

  Bugfix (non-breaking change which fixes an issue)

### DIFF
--- a/RestSharp/Parameter.cs
+++ b/RestSharp/Parameter.cs
@@ -27,8 +27,11 @@ namespace RestSharp
     {
         public Parameter(string name, object value, ParameterType type)
         {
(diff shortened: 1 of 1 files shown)


mitcht:
I am used to bitbucket.  Sorry about reopening repeatedly.

mitcht:
Not sure how the checks are failing.

alexeyzimarev:
It has been fixed in another PR, it was indeed a big issue, sorry.

Issue #1304, opened 2019-04-09
Updating Authenticators


mattosaurus:
## Expected Behavior
Authenticators should update request parameter value if name already exists.

## Actual Behavior
Authenticate method checks for name and only adds parameter if it doesn't exist.

## Steps to Reproduce the Problem

  1. Add a new client with a JwtAuthenticator with an expired token.
  2. Execute request and receive expired 401 error.
  3. Fetch new token and set client authenticator to be new JwtAuthenticator with valid token.
  4. Execute request again with updated client.

I'm trying to implement an execute method using the JwtAuthenticator that will acquire a new token if the current one is expired and then try to send the request again. However this new token isn't being set in the request parameters as a parameter with the name Authorization already exists.

I think the expected behavior should be that if a new authenticator is added then it will update the existing parameter rather than doing nothing.

This should hopefully just be a case of using the AddOrUpdateParameter method rather than the AddParameter method in the authenticators.

`request.AddParameter("Authorization", authHeader, ParameterType.HttpHeader);`

to

`request.AddOrUpdateParameter("Authorization", authHeader, ParameterType.HttpHeader);`

I'm happy to submit a PR with these changes if you agree with this.

Thanks

alexeyzimarev:
Ok, I made some changes to your code. I want to keep the behaviour as it was but give the possibility to address the issue with supplying a new token.

Now, there're two ways to update the token.

1. Use the new constructor: `new JwtAuthenticator(token, true)`, where the second parameter tells the client to update the header parameter. If the second parameter is `false` or not supplied (which is the default), it will work as before.

1. Keep the authenticator reference in a variable and call `authenticator.SetBearerToken(token)`, which will also force the header update.

PR #1338, opened 2019-07-22
Fixes issue https://github.com/restsharp/RestSharp/issues/1304

mitcht:
## Description

This pull request fixes issue https://github.com/restsharp/RestSharp/issues/1304

## Purpose
This pull request is a:

- [x ] Bugfix (non-breaking change which fixes an issue)

### DIFF
--- a/RestSharp/Authenticators/JwtAuthenticator.cs
+++ b/RestSharp/Authenticators/JwtAuthenticator.cs
@@ -40,10 +40,7 @@ namespace RestSharp.Authenticators
 
         public void Authenticate(IRestClient client, IRestRequest request)
         {
-            // only add the Authorization parameter if it hasn't been added by a previous Execute
-            if (!request.Parameters.Any(p => p.Type.Equals(ParameterType.HttpHeader) &&
-                                             p.Name.Equals("Authorization", StringComparison.OrdinalIgnoreCase)))
-                request.AddParameter("Authorization", authHeader, ParameterType.HttpHeader);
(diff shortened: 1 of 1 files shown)


mitcht:
This test fails because it is wrong.  You should not be allowed to have two headers with the same name, regardless if they are different types.

This authentication method in the release allows a querystring header with a type of `"QueryString"` and `Name == "Authorization"` to simultaneously exist with a parameter of type `"HttpHeader"` with a `Name == "Authorization"` and ContentType of Bearer .....

https://github.com/restsharp/RestSharp/blob/a8a9f3473be30317fc449c1c47e3f3cd74e8bab4/RestSharp/Authenticators/JwtAuthenticator.cs#L41-L47

The following test `Check_Only_Header_Authorization` in JwtAuthTests.cs looks for two parameters, both of which are Authorization type.  This seems wrong.  Please advise.

https://github.com/restsharp/RestSharp/blob/a8a9f3473be30317fc449c1c47e3f3cd74e8bab4/RestSharp.Tests/JwtAuthTests.cs#L66-L68

alexeyzimarev:
Thank you for the PR. I need to check those tests but right now we have issues with AppVeyor and I am planning to move out from there instead of fighting with it.

alexeyzimarev:
@mitcht have you read the comment in that test?

```
            // Paranoid server needs "two-factor authentication": jwt header and query param key for example
            request.AddParameter("Authorization", "manualAuth", ParameterType.QueryString);
```

I have never experienced such an issue myself but it seems like someone did. There are no two identical header parameters, one parameter is the header parameter and another one is the query string parameter.

PR #1408, opened 2020-01-16
Throw exception on invalid resource query

### LINKED ISSUE #1404: Value cannot be null error on RestRequest
## Steps to Reproduce the Problem

  1. Use version 106.10.1
  2. Try this code    var request = new RestRequest(uri.AbsoluteUri);
  3. This version works 106.6.10.

## Specifications

  - Version: 106.10.1
  - Platform:
  - Subsystem:

## StackTrace
<details>

System.ArgumentNullException: Value cannot be null.
Parameter name: name
  at RestSharp.Validation.Ensure.NotEmpty (System.String parameter, System.String name) [0x00008] in <fba61e8429484999a16cc36b262a4d33>:0
  at RestSharp.Parameter..ctor (System.String name, System.Object value, RestSharp.ParameterType type) [0x00011] in <fba61e8429484999a16cc36b262a4d33>:0
  at RestSharp.RestRequest.AddParameter (System.String name, System.Object value, RestSharp.ParameterType type) [0x00000] in <fba61e8429484999a16cc36b262a4d33>:0
  at RestSharp.RestRequest.AddQueryParameter (System.String name, System.String value, System.Boolean encode) [0x00000] in <fba61e8429484999a16cc36b262a4d33>:0
  at RestSharp.RestRequest..ctor (System.String resource, RestSharp.Method method, RestSharp.DataFormat dataFormat) [0x00088] in <fba61e8429484999a16cc36b262a4d33>:0
  at RestSharp.RestRequest..ctor (System.String resource) [0x00000] in <fba61e8429484999a16cc36b262a4d33>:0
  at SNLA.Core.Communication.SNLAHttpClient.GetRaw (System.Uri uri, System.Boolean authenticated) [0x00023] in /Users/jorgenstorlie/Apper/SNLA-app/SNLA.Core/Communication/SNLAHttpClient.cs:182
  at SNLA.Core.Communication.VianettSMSService.CallVianettSMS (System.String vianettBaseUrl, System.Net.NetworkCredential vianettCredential, System.String mobileNumber, System.String message, SNLA.Core.Communication.VianettSMSService+SenderAddressType senderAddressType, SNLA.Core.Communication.VianettSMSService+Priority priority) [0x00064] in /Users/jorgenstorlie/Apper/SNLA-app/SNLA.Core/Communication/VianettSMSService.cs:44
  at Hjelp113.iOS.OnboardingMobileViewController.SendPhoneNumber (System.String phoneNumber) [0x000af] in /Users/jorgenstorlie/Apper/113Xamarin/Hjelp113.iOS/OnboardingMobileViewController.cs:147
  at Hjelp113.iOS.OnboardingMobileViewController.Register () [0x000b1] in /Users/jorgenstorlie/Apper/113Xamarin/Hjelp113.iOS/OnboardingMobileViewController.cs:114
  at Hjelp113.iOS.OnboardingMobileViewController.BtnDone_TouchUpInside (System.Object sender, System.EventArgs e) [0x00032] in /Users/jorgenstorlie/Apper/113Xamarin/Hjelp113.iOS/OnboardingMobileViewController.cs:63
  at System.Runtime.CompilerServices.AsyncMethodBuilderCore+<>c.<ThrowAsync>b__7_0 (System.Object state) [0x00000] in /Library/Frameworks/Xamarin.iOS.framework/Versions/Current/src/Xamarin.iOS/mcs/class/referencesource/mscorlib/system/runtime/compilerservices/AsyncMethodBuilder.cs:1021
  at Foundation.NSAsyncSynchronizationContextDispatcher.Apply () [0x00000] in /Library/Frameworks/Xamarin.iOS.framework/Versions/13.8.3.0/src/Xamarin.iOS/Foundation/NSAction.cs:178
  at at (wrapper managed-to-native) UIKit.UIApplication.UIApplicationMain(int,string[],intptr,intptr)
  at UIKit.UIApplication.Main (System.String[] args, System.IntPtr principal, System.IntPtr delegate) [0x00005] in /Library/Frameworks/Xamarin.iOS.framework/Versions/13.8.3.0/src/Xamarin.iOS/UIKit/UIApplication.cs:86
  at UIKit.UIApplication.Main (System.String[] args, System.String principalClassName, System.String delegateClassName) [0x0000e] in /Library/Frameworks/Xamarin.iOS.framework/Versions/13.8.3.0/src/Xamarin.iOS/UIKit/UIApplication.cs:65
  at Hjelp113.iOS.Application.Main (System.String[] args) [0x00001] in /Users/jorgenstorlie/Apper/113Xamarin/Hjelp113.iOS/Main.cs:12

</details>

Deantwo:
## Description

Throw exception on invalid resource query
Fixes #1404

## Purpose
This pull request is a:

## Checklist

### DIFF
--- a/src/RestSharp/RestRequest.cs
+++ b/src/RestSharp/RestRequest.cs
@@ -92,6 +92,11 @@ namespace RestSharp
                     .Select(
                         x =>
                         {
+                            if (x == string.Empty)
+                                throw new ArgumentException(
+                                    "resource contains an invalid query."
+                                );
+                            
                             var position = x.IndexOf('=');
 
                             return position > 0


Deantwo:
Old behavior mentioned in #1404 allowed a query with `?&`. I don't think this is allowed in a Uri, is it?
If not then we can simply change it to ignore empty entries.
``` C#
query.Split(new char[] { '&' }, StringSplitOptions.RemoveEmptyEntries);
```
But I would think that throwing an error to make the user aware that the Uri is wrong is a good idea.
Thoughts?

Deantwo:
Found this nice little article about query strings, and it sadly doesn't answer the question of rather it is legal or not to have empty key and null value, but it does suggest that it at least makes no sense in JSON.

I looked at the RFC, but it didn't seem to mention any finer rules on this that I could see.
I mean need to know if these are valid and should be ignored, or if an ArgumentException should be thrown.
Examples:
- `?&`
- `&&`
- `&`(end of string)

Deantwo:
```
test?key1=value1&=value2
```
Result in:
image
This pull request doesn't affect this.

So it feels like the entire `static IEnumerable<NameValuePair> ParseQuery(string query)` method might need to be changed. Something I don't know if i can do, or if it is even a good idea to do it in the first place.
Might be better to find some built-in features or another library to do this correctly, rather than try to re-invent the wheel again.

alexeyzimarev:
The thing is that the `Uri` class constructor does some checks to ensure the given uri is valid. Apparently, it doesn't see an issue with `?&`. I think ignoring empty entries is the best way to handle this. I don't think throwing an exception is a good idea since it will possibly introduce a breaking change.

PR #1414, opened 2020-01-24
Ignore Empty Uri Query Parameters

Deantwo:
## Description

A different attempt at fixing #1404 then what was proposed in #1408.
But even this is not a perfect solution, as explained in https://github.com/restsharp/RestSharp/pull/1408#issuecomment-575213217 and https://github.com/restsharp/RestSharp/pull/1408#issuecomment-575596890.

Someone with more knowhow should probably rewrite the entire [`ParseQuery(string)`](https://github.com/restsharp/RestSharp/blob/master/src/RestSharp/RestRequest.cs#L83) method. Or even better find a more standard way of doing it that doesn't require us to maintain the function ourselves.

For a good idea about how the query parsing should handle can be seen in this little article: https://zzzzbov.com/blag/query-string-hell

## Purpose
This pull request is a: Bugfix (non-breaking change which fixes an issue)

## Checklist

### DIFF
--- a/src/RestSharp/RestRequest.cs
+++ b/src/RestSharp/RestRequest.cs
@@ -88,7 +88,7 @@ namespace RestSharp
             }
 
             static IEnumerable<NameValuePair> ParseQuery(string query)
-                => query.Split('&')
+                => query.Split(new char[] { '&' }, StringSplitOptions.RemoveEmptyEntries)
                     .Select(
                         x =>
                         {


Deantwo:
> what is the motivation to provide an array with one element instead of using just a `char` as we did before?

The [`system.string.split`](https://docs.microsoft.com/en-us/dotnet/api/system.string.split?view=netframework-4.8) method only has one overload with `params char[]` as parameter. If you want to use any `StringSplitOptions` options, you have to supply it as a `char[]` array.

alexeyzimarev:
I guess we need a test for this failing case too?

Deantwo:
> I guess we need a test for this failing case too?

As mentioned in https://github.com/restsharp/RestSharp/pull/1408#issuecomment-575596890, there are still a few issues that this doesn't solve. But yes it would probably be a good idea to make some test cases based on the examples given in the https://zzzzbov.com/blag/query-string-hell article.

But it might be even better if the [`ParseQuery(string)`](https://github.com/restsharp/RestSharp/blob/master/src/RestSharp/RestRequest.cs#L83) method could be replaced with some built-in function or something. But I haven't had time to research that while at work, since fixing open-source bugs isn't all that good with my boss looking over my shoulder.

Issue #1452, opened 2020-04-28
Add property name to the FormatException


andrasfuchs:
I have the following XML response from an API call:
```
<?xml version='1.0' encoding='UTF-8'?>
<queryresult success='true'
    error='false'
    numpods='1'
    datatypes=''
    timedout=''
    timedoutpods=''
    timing='1.0230000000000001'
    parsetiming='0.613'
    parsetimedout='false'
    recalculate=''
    id='MSP1801h5c3fg92f0194960000253b8ice3ghfff5i'
    host='https://www5b.wolframalpha.com'
    server='31'
    related='https://www5b.wolframalpha.com/api/v1/relatedQueries.jsp?id=MSPa1811h5c3fg92f01949600001di79bh0edie2i4e5282741642789771482'
    version='2.6'>
 <pod title='Solution'
     scanner='Reduce'
     id='Solution'
     position='100'
     error='false'
     numsubpods='1'
     primary='true'>
  <subpod title=''>
   <img src='https://www5b.wolframalpha.com/Calculate/MSP/MSP1821h5c3fg92f01949600000edggfa59013habb?MSPStoreType=image/gif&amp;s=31'
       alt='x = -1'
       title='x = -1'
       width='550'
       height='22'
       type='Default'
       themes='1,2,3,4,5,6,7,8,9,10,11,12'
       colorinvertable='true' />
   <plaintext>x = -1</plaintext>
  </subpod>
  <expressiontypes count='1'>
   <expressiontype name='Default' />
  </expressiontypes>
  <states count='1'>
   <state name='Step-by-step solution'
       input='Solution__Step-by-step solution'
       stepbystep='true' />
  </states>
 </pod>
</queryresult>
```

During deserialization RestSharp throws a FormatException while it tries to deserialize the "timing" property. 

The exact Exception looks like this:
```
Exception thrown: 'System.FormatException' in System.Private.CoreLib.dll
An exception of type 'System.FormatException' occurred in System.Private.CoreLib.dll but was not handled in user code
Input string was not in a correct format.
```

My request is to include information about the problematic property in the Exception message. The Exception is generated in the `protected virtual object Map(object x, XElement root)` method of the `XmlDeserializer` class. The `Property` object is available there, so the `prop.Name`, `prop.Type` could be included fairly easily.

## Actual Behavior

There is no detailed information about the reason why the exception was thrown. It's really difficult to debug the issue, if we don't know where to look in the XML. 

## Steps to Reproduce the Problem

[...]

PR #1462, opened 2020-05-12
Add property name to the FormatException #1452

adityamagadi:
Fixed format exception error

## Description

Add property name to the FormatException

## Purpose
This pull request is a:

## Checklist

### DIFF
--- a/src/RestSharp/Serializers/Xml/XmlDeserializer.cs
+++ b/src/RestSharp/Serializers/Xml/XmlDeserializer.cs
@@ -188,7 +188,14 @@ namespace RestSharp.Deserializers
                 }
                 else if (type.IsPrimitive)
                 {
-                    prop.SetValue(x, value.ChangeType(asType), null);
+                    try
+                    {
+                        prop.SetValue(x, value.ChangeType(asType), null);
+                    }
+                    catch (FormatException)
+                    {
+                        throw new FormatException("Format exception while Deserializing Property Name - " + prop.Name);
+                    }
                 }
                 else if (type.IsEnum)
                 {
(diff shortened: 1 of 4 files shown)


src/RestSharp/Serializers/Xml/XmlDeserializer.cs
```diff
@@ -188,7 +188,14 @@ protected virtual object Map(object x, XElement root)
...
                 else if (type.IsPrimitive)
                 {
-                    prop.SetValue(x, value.ChangeType(asType), null);
+                    try
+                    {
+                        prop.SetValue(x, value.ChangeType(asType), null);
+                    }
+                    catch (FormatException)
+                    {
+                        throw new FormatException("Format exception while Deserializing Property Name - " + prop.Name);
```
andrasfuchs, line 197: Could you include the original FormatException as InnerException of the new one?

Also could you add a little more detail to the message itself. I would imagine something like this:
$"Couldn't parse the value of '{value}' into the '{prop.Name}' property, because it isn't a type of '{prop.Type}'."

adityamagadi: Thanks for the inputs, I have taken care of these changes. Can you please review the same?

adityamagadi:
> Excellent

Thank you. Please let me know when the fix is merged to the main repo.

andrasfuchs:
> Thank you. Please let me know when the fix is merged to the main repo.

You will get a notification via e-mail. 
The owner(s) of the RestSharp repo must approve your pull-request too, I'm not one of them :)

alexeyzimarev:
> @alexeyzimarev, could you take a look?

It's my issue, not even sure how it passed the test before :/

alexeyzimarev:
Thank you both for contributing :)

PR #1464, opened 2020-05-18
Use random value for multipart/form-data boundary.

Trellmor:
## Description

The request body might contain the previous, fixed value for FormBoundary.
This will cause a server to split the body on the wrong boundaries.

For example trying to post this file which already contains a boundary will generate an invalid request.

`-------------------------------28947758029299
Content-Disposition: form-data; name="somefile_REC.pdf"; filename="somefile_REC.pdf"
Content-Type: application/x-gzip

%PDF-1.4
[...] truncated
%%EOF

-------------------------------28947758029299--
`

By using a random value for the boundary collisions like this will be avoided.

## Purpose
This pull request is a:

## Checklist

### DIFF
--- a/src/RestSharp/Http.cs
+++ b/src/RestSharp/Http.cs
@@ -34,7 +34,7 @@ namespace RestSharp
     {
         const string LineBreak = "\r\n";
 
-        const string FormBoundary = "-----------------------------28947758029299";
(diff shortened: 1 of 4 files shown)


alexeyzimarev:
Cool, thanks!

PR #1476, opened 2020-06-23
Update JsonNetSerializer.cs

KitKat31337:
This should not default to a non-standard contract resolver that ignores standard newtonsoft JSON.net attributes and causes much  confusion with many APIs where case matters.

The DefaultSettings should initialize with the DefaultContractResolver

## Purpose
This pull request is a:

This is a breaking change if someone used the library and instead of realizing the non-standard resolver is defaulted, adjusted their business process instead

## Checklist


alexeyzimarev:
It's a breaking change, as you can see in the test. Camel case is the de-facto default in JSON world and I'm afraid I indent to keep it as is. It's easy to supply the default contract resolver using the custom settings supplied to the serialiser.

Issue #1491, opened 2020-09-04
Json deserialization is not working as expected when culture is 'az' or 'tr'


nesc58:
## Expected Behavior
Deserialized objects should be equivalent independend of `Thread.CurrentThread.CurrentUICulture` or `Thread.CurrentThread.CurrentCulture`.

## Actual Behavior
When the culture is `az` or `tr` or a derived culture of them, serialization not working for properties starting with uppercase `I`.

In the following example the deserialized object contains `null` instead of `MyName`.

DTO:
```
public class Person
{
    public string InternalName {get; set;}
}
```

Json:
```
{
    "internalName": "MyName"
}
```

## Steps to Reproduce the Problem
Here is a repo with unit tests to test all cultures:
https://github.com/nesc58/RestSharpCultureTests

## Specifications

  - Version: `106.11.4`
  - Platform: Windows
  - Subsystem: Windows 10, .netcoreapp 2.2

I am not able to find the right place to fix this. I have found some interesting code places:
While mapping the properties, `TryGetValue` returns `null`: https://github.com/restsharp/RestSharp/blob/dev/src/RestSharp/Serializers/Json/JsonSerializer.cs#L120

It might be better not to use the GenericEqualityComparer<string> (default one): https://github.com/restsharp/RestSharp/blob/dev/src/RestSharp/Serializers/Json/SimpleJson.cs#L688

Perhaps the problem is solved by using a better (culture invariant comparer) like `StringComparer.InvariantCultureIgnoreCase` or `StringComparer.InvariantCulture`.

Initialization could be easy done by: `new JsonObject(StringComparer.InvariantCulture)`

nesc58:
I have found something interesting:
`var variants = parts[i].GetNameVariants(Culture).Distinct();`
returns for `InternalName` e.g. `ınternalName`  instead of `internalName`

This is not the expected character

alexeyzimarev:
Yeah, seems like the issue is indeed in culture settings. That's why the issue is only experienced by some and not all users. It might even not be found by tests.

nesc58:
I have created a pull request: https://github.com/restsharp/RestSharp/pull/1492

nesc58:
@alexeyzimarev is there a planned release date for `v106.11.5`? What are the open issues or features to release this fix?

alexeyzimarev:
I want to look at two recent issues #1495 #1497 and try to fix them before releasing the new version.
Your fix is in the latest preview package already.

[...]

PR #1492, opened 2020-09-04
Fix culture

nesc58:
## Description

Use given culture when call `.ToLower`

The function `MakeInitialLowerCase` have breaking change. We can decide to use `CultureInfo.InvariantCulture` as default fallback.

Fix for Issue: https://github.com/restsharp/RestSharp/issues/1491

## Purpose
This pull request is a:

- [ x ] Bugfix (non-breaking change which fixes an issue)
- [ x ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist

- [ x ] I have added tests that prove my fix is effective or that my feature works


alexeyzimarev:
I think it's ok to change the signature for those functions. They should be, in fact, internal. It's not for RestSharp to provide string conversion tools. Thanks for the PR!

PR #1498, opened 2020-09-11
Encode oauth_token. Fixes #1495

### LINKED ISSUE #1495: OAuth failing because oauth_token is not encoded.
## Expected Behavior
OAuth authenticator should get access token.

## Actual Behavior
Using `OAuth1Authenticator.ForAccessToken` I get an error from the provider that `oauth_token` is absent.  But it exists in the header. 
After debugging the source code, I figured out that `RestSharp` doesn't encode `oauth_token`.

After changing the line in `OAuthWorkflow` file:
 `if (!Token.IsEmpty()) authParameters.Add(new WebPair("oauth_token", Token)); `

to

` if (!Token.IsEmpty()) authParameters.Add(new WebPair("oauth_token", Token, true));`

I was managed to get access token.

PS. I use ETrade exchange broker in this case.

## Specifications

  - Version: 106.11.4 ( also tried 106.11.5-alpha.0.21)
  - Platform: .NET Core 3.1
  - Subsystem:  N/A

sharok:
## Description
In reference to #1495 
 
According to RFC, the `oauth_token` must be encoded.  Some OAuth providers return token that contains `=` symbol, the current implementation doesn't encode it to %3D.  This PR adding encoding for `oauth_token`.

## Purpose
This pull request is a:

## Checklist

### DIFF
--- a/src/RestSharp/Authenticators/OAuth/OAuthWorkflow.cs
+++ b/src/RestSharp/Authenticators/OAuth/OAuthWorkflow.cs
@@ -10,7 +10,7 @@
 //   distributed under the License is distributed on an "AS IS" BASIS,
 //   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 //   See the License for the specific language governing permissions and
-//   limitations under the License. 
+//   limitations under the License.
 
 using System;
 using System.Collections.Generic;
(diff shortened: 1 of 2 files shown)


alexeyzimarev:
Thanks for the PR! Is there a way to add a test, which would prevent this issue from popping up in the future?

sharok:
@alexeyzimarev  just added the test.

Issue #1501, opened 2020-09-18
Deprecation warnings using Parameter outside of IRestRequest


nesc58:
## Expected Behavior
No deprecation warnings.

## Actual Behavior
Writing custom `IRestSerializer` implementation produces some deprecation warning
Deprecation warning: `"Use Add[XXX]Parameter methods of IRestRequest instead of instantiating the Parameter class."`

# General / Questions
Parameters seems not be deprecated at all.
RestSharp self still uses the `Parameter` class e.g. in `IRestRequest`: `List<Parameter> Parameters { get; }`.
Will `Parameter` class removed / replaced at some point? 

# Suggestion
* Remove deprecation warnings for `Parameter` class.
* Add deprecation warnings to `IRestRequest`
  * `IRestRequest AddParameter(Parameter p)`
  * `IRestRequest AddOrUpdateParameter(Parameter parameter);`
  * Maybe: `AddOrUpdateParameters(IEnumerable<Parameter> parameters)`

alexeyzimarev:
The message is not correctly placed, as you mentioned the list of parameters will be available.

However, I want to make parameters more robust by encapsulating more.
- It should not be a `List` but `IReadOnlyCollection` so it cannot be modified
- Parameters to be added only by using `AddParameter` method or any new API.
- Make the `Parameter` class constructor internal

However, I might see how it will hinder some of the use cases and I'd like to hear some feedback.

alexeyzimarev:
Ideally, I'd keep the `Parameter` class internal but I am not sure how people use this collection to make a decision.

nesc58:
I have seen some code lines where instances of `Paramter` written to class fields because they were reused a lot of times. e.g. generating `Authorized` headers instead of using an _Authenticator_. Sometime authentication is only required for special routes and not at all. 

What do you think? Is there an alternative for that? In my mind: Creating multiple instances solves the problem: One with authenticator, one without. But this seems a little bit messy.

Slushnas:
I was confused by this message as well.

I am using a custom helper class to construct a list of parameters based on some conditions and then I pass that list to the `AddOrUpdateParameters()` method. Is there a downside in allowing us to construct a list of parameters outside one of the Add[XXX]Parameter methods?

alexeyzimarev:
That's basically why I need feedback. 👍

alexeyzimarev:
The reason for encapsulation is this:
[...]

PR #1503, opened 2020-09-18
[DRAFT] Update Parameters handling

nesc58:
## Description
Issue: https://github.com/restsharp/RestSharp/issues/1501

To permit adding/removing Parameters from _outside_, I have created a custom readonly collection. 

Internally adding/removing elements (required for `Add[XXX]Parameter` methods) is still possible.

Modifying `Parameters` directly via the following is not longer possible
```
var request = new RestRequest();
request.Parameters.Add(new Parameter())
```

## Purpose
This pull request is a:

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist


alexeyzimarev:
I don't think we need a new class. We can keep private collection as `List` and expose it via `IReacOnlyCollection<Paramerer> Parameters => _parameters.AsReadOnlyCollection()'

nesc58:
Yes, you're absolutely right. It was a little bit too much.

In my mind it was something like:
* Not changing API too much
* It would be nice to have a single property.
* It would also be nice to have some methods like `FindAll` or array notation `[]` directly on it (was used in some tests)

But you're right. Having a private List<T> field and a getter property which wraps the field into ReadOnlyCollection<T> would also solve it (`Parameters =_parameters.AsReadOnly()`)

Another point we must discuss:
Should Parametes always reflect the inner List?
e.g.
```
var request = new RestRequest().AddQueryParameter("one", "one");
var firstParameters = request.Parameters;
request = request.AddQueryParameter("two", "two");
var secondParameters = request.Parameters;
```

Should the number of firstParameters and secondParameters identically?

The default behaviour is: firstParameters and secondParameters are identically, because AsReadOnly() is only a wrapper rather than a copied read-only version

alexeyzimarev:
Right now it works by direct reference. I believe `AsReadOnly` just does the cast so the collection would still be the same reference. It's not an immutable list.

It will be a big change when people find out that they can't just drop everything they want to the parameters collection. I first want to finalise the decision about the `Parameter` abstraction level before doing anything with the collection.

nesc58:
>  I first want to finalise the decision about the Parameter abstraction level before doing anything with the collection.


I think this is the right way to do it.

nesc58:
It is only a draft,
A lot of things are not implemented / documented. I have tried to create abstract classes for the different parameter types.

For now, I am not happy with the IRestClient / IRestRequest interface. There are a lot of breaking changes.
e.g. `AddParameter(string, object, ParameterType)` are not existing anymore.

It would be better to have defined methods for each parameter instead of a generic AddParameter(string, object, type). E.g. the value of a query string parameter should be a string; similiar HttpHeaders, UrlSegments and so on.

Maybe we could provide something compatible like:
```
public IRestRequest AddParameter(string name, object value, ParameterType type)
{
	switch (type)
	{
		case ParameterType.GetOrPost:
			return AddParameter(name, value);
		case ParameterType.UrlSegment:
			return AddUrlSegment(name, value);
		case ParameterType.HttpHeader:
			return AddHeader(name, (string)value);
		case ParameterType.QueryString:
			return AddParameter(new QueryStringParameter(name, value));
		case ParameterType.QueryStringWithoutEncode:
			return AddParameter(new QueryStringParameter(name, value, false));
		default:                                     throw new ArgumentOutOfRangeException(nameof(type), type, null);
	}
}
```

The thing i struggle with is: I think a RequestBody must have a content type. So creating it without a content type will result in something unexpecting.

In the next days I want to analyse the default behaviour of the current implementations to determine a set of default values for the paramters.

alexeyzimarev:
The PR actually looks good. The only issue I see is that the `Parameter` class itself should become abstract but it's impossible to do without breaking changes. But we can do it in 107.

nesc58:
Hi,
there are some breaking changes:
IRestRequest.Parameters changed to IReadOnlyCollection
IRestRequest.Files changed to IReadOnlyCollection
Parameter changed to abstract class but there is a ParameterFactory which can create parameters of specific ParameterTypes.

Issue #776, opened 2015-11-02
OAuth 1.0 auth not working for empty consumer secret


ivan-milovanovic-troxo:
I tried to use RestSharper as client for [MAAS RestAPI](https://maas.ubuntu.com/docs/api_authentication.html) and stumbled upon an issue with OAuth 1.0 authentication implementation. 

RestSharper does not allow empty consumer secret, this is incorrect cause consumer secret can be empty string. "The Consumer Secret MAY be an empty string" from [OAuth Core 1.0](http://oauth.net/core/1.0a/)

In [OAuthWorkflow](https://github.com/restsharp/RestSharp/blob/master/RestSharp/Authenticators/OAuth/OAuthWorkflow.cs#L296-L299) exception is being thrown if consumer secret is empty:

```
    if (this.ConsumerSecret.IsNullOrBlank())
    {
        throw new ArgumentException("You must specify a consumer secret");
    }
```

hallem:
No problem that's an easy fix.  I'll add it to the list.

PR #1596, opened 2021-06-13
OAuth1: Allow empty customer secrets, according to the specifications.

JamesMenetrey:
## Description

OAuth 1.0a specifications allow to use of an empty custom secret (see [here](https://oauth.net/core/1.0a/, section 4: _The Consumer Secret MAY be an empty string [..]_). Similarly to #776, I needed to call MAAS API, where the customer secret must not be provided (0-legged OAuth). This pull request removes the constraint of providing customer secrets for OAuth1.

This has been tested over the API of MAAS, which does not require any customer secret.

## Purpose
This pull request is a:

## Checklist

- [X] I have added tests that prove my fix is effective or that my feature works (added one to verify that the customer secret can be empty).
- [X] I have added necessary documentation (explained the purpose of this PR on a comment over the provided unit test).

Don't hesitate to ask for changes in that pull request if needed.

Cheers

### DIFF
--- a/src/RestSharp/Authenticators/OAuth/OAuth1Authenticator.cs
+++ b/src/RestSharp/Authenticators/OAuth/OAuth1Authenticator.cs
@@ -12,15 +12,13 @@
 //   See the License for the specific language governing permissions and
 //   limitations under the License. 
 
+using JetBrains.Annotations;
+using RestSharp.Authenticators.OAuth;
+using RestSharp.Extensions;
 using System;
 using System.Collections.Generic;
 using System.Linq;
-using System.Text;
 using System.Web;
-using JetBrains.Annotations;
-using RestSharp.Authenticators.OAuth;
-using RestSharp.Authenticators.OAuth.Extensions;
-using RestSharp.Extensions;
 
 // ReSharper disable CheckNamespace
 
@@ -42,7 +40,7 @@ namespace RestSharp.Authenticators
 
         internal virtual string ConsumerKey { get; set; }
 
-        internal virtual string ConsumerSecret { get; set; }
+        internal virtual string? ConsumerSecret { get; set; }
 
         internal virtual string Token { get; set; }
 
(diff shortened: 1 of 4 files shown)


patrickhampson:
Hey James,

Funny, I stumbled upon this as I was opening my own PR for this exact MAAS api issue!

I opened #1597 explaining the behavior a bit but wanted to comment on this PR rather than opening a duplicate:

- I recommend leaving the OAuth consumerSecret parameter as a regular string vs the nullable type you changed it to.  This should reduce the impact of this change as it would be a signature change vs an internal validation change.

- It might also be beneficial to modify the unit test to account for the different OAuthTypes:
```
[Test]
[TestCase(OAuthType.AccessToken)]
[TestCase(OAuthType.ProtectedResource)]
[TestCase(OAuthType.AccessToken)]
[TestCase(OAuthType.ProtectedResource)]
public void Authenticate_ShouldAllowEmptyConsumerSecret_OnHttpAuthorizationHeaderHandling(OAuthType type)
{
    // Arrange
    const string url = "https://no-query.string";

    var client = new RestClient(url);
    var request = new RestRequest();
    _authenticator.Type           = type;
    _authenticator.ConsumerSecret = "";

    // Act
    _authenticator.Authenticate(client, request);

    // Assert
    var authParameter = request.Parameters.Single(x => x.Name == "Authorization");
    var value = (string)authParameter.Value;

    Assert.IsNotEmpty(value);
    Assert.IsTrue(value!.Contains("OAuth"));
    Assert.IsTrue(value.Contains("oauth_signature=\"" + OAuthTools.UrlEncodeStrict("&")));
}

JamesMenetrey:
Hello Patrick,

Many thanks for your feedback and I am happy to see some tractions around this use case! I have applied your suggestion regarding the enhancement of the unit test.

> I recommend leaving the OAuth consumerSecret parameter as a regular string vs the nullable type you changed it to. This should reduce the impact of this change as it would be a signature change vs an internal validation change.

I hesitated before changing these signatures. My reasoning was that introducing a nullable type for the parameter `consumerSecret` would be ok, since C# allowed an implicit cast from `string?` to `string` and this conveys the intent that the consumer secret may be empty. In the case I restore the initial type of the parameter (`string`), would you still allow passing `null` as an argument to `consumerSecret`, or only allow an empty string to be passed? Should we wait the feedback of a RestSharp's dev before changing it?

Cheers!

PR #1619, opened 2021-07-26
Better enum deserialization exception fixes #1614

### LINKED ISSUE #1614: Enum deserialization exception is not very helpful
## Expected Behavior
For a response where the following class is serialized:
```
public class Config
{
    public List<Option> Options { get; set; }
}
```

With server-side code:
```
public enum Option
{
    Zero,
    One,
    Two,
    Three
};
```

And a client-side bug:
```
public enum Option
{
    Zero,
    One,
    Two
    // Three is missing
};
```

I would hope to get a meaningful exception.

## Actual Behavior
The exception tells me about the internals of RestSharp, not about the type being deserialized:
```
Value cannot be null.
Parameter name: item
```

## Specifications
  - Version: Any recent version including 106.12.0 and dev branch
  - Platform: Windows 10

## StackTrace
```
System.ArgumentNullException
   at System.ThrowHelper.IfNullAndNullsAreIllegalThenThrow[T](Object value, ExceptionArgument argName)
   at System.Collections.Generic.List`1.System.Collections.IList.Add(Object item)
   at RestSharp.Serialization.Json.JsonSerializer.BuildList(Type type, Object parent)
   at RestSharp.Serialization.Json.JsonSerializer.ConvertValue(TypeInfo typeInfo, Object value)
   at RestSharp.Serialization.Json.JsonSerializer.Map(Object target, IDictionary`2 data)
   at RestSharp.Serialization.Json.JsonSerializer.CreateAndMap(Type type, Object element)
   at RestSharp.Serialization.Json.JsonSerializer.ConvertValue(TypeInfo typeInfo, Object value)
   at RestSharp.Serialization.Json.JsonSerializer.Deserialize[T](IRestResponse response)
   at RestSharp.RestClient.Deserialize[T](IRestRequest request, IRestResponse raw)
```

So the code path is:
\RestSharp\src\RestSharp\Serializers\Json\JsonSerializer.cs
```
    IList BuildList(Type type, object? parent) {
    // ...
                    var item = ConvertValue(itemType.GetTypeInfo(), element);

                    list.Add(item); // ** Thrown here because item is null. **
```
\RestSharp\src\RestSharp\Serializers\Json\JsonSerializer.cs
```
    object? ConvertValue(TypeInfo typeInfo, object? value) {
    // ...
        if (typeInfo.IsEnum) return type.FindEnumValue(stringValue, Culture);
```
```
\RestSharp\src\RestSharp\Extensions\ReflectionExtensions.cs
    public static object? FindEnumValue(this Type type, string value, CultureInfo culture)
    // ...
        return ret; // ** Returns the null which will cause the exception **
```

## Possible fix

It would be nice to return information that the caller could use to identify the problem. Different info is available at different levels in the stack, and devs more familiar with the RestSharp codebase could craft something elegant. But the simplest change would seem to be something like a null check at the end of FindEnumValue():
```
    if (ret == null)
        throw new ArgumentOutOfRangeException(value + " undefined for " + type);
    return ret;
```
Which gives an exception message such as:
```
Exception: Specified argument was out of the range of valid values.
Parameter name: 3 undefined for Option
```
I can submit this as a PR if desired, though it's trivial and you'll likely want to change something.

sicklittlemonkey:
Avoid null exception, and return better error info for client/server enum mismatch.

## Description

This simple suggestion fixes #1614.

## Purpose
This pull request is a:

## Checklist

(I had a quick look, but didn't see a similar test that I could re-use for this case.)

### DIFF
--- a/src/RestSharp/Extensions/ReflectionExtensions.cs
+++ b/src/RestSharp/Extensions/ReflectionExtensions.cs
@@ -93,6 +93,8 @@ namespace RestSharp.Extensions
             if (enumValueAsUnderlyingType != null && Enum.IsDefined(type, enumValueAsUnderlyingType))
                 ret = (Enum) Enum.ToObject(type, enumValueAsUnderlyingType);
 
+            if (ret == null)
+                throw new ArgumentOutOfRangeException(value + " undefined for " + type);
             return ret;
         }
     }


alexeyzimarev:
@sicklittlemonkey The test `Deserialization_Of_Undefined_Int_Value_Returns_Enum_Default` started to fail after the fix, so it's a breaking change. Have you executed the test suite after you made the change? It was returning the "default" value, which is `null`.

alexeyzimarev:
I removed the trow and made it return the default value of a given type. It means that it won't throw the argument out of range exception, but it also won't fail with null reference exception. However, you'd not be aware that the enum doesn't contain the value it must have.

I am not sure how it works with something like `System.Text.Json`, it might just return `null` for the whole REST model. At least it is doing so by default when you, for example, put an invalid date to the `DateTime` field. Overall, I plan to kill SimpleJson and use `System.Text.Json` as a default serialiser since it's embedded in all the latest .NET versions anyway. The reason for SimpleJson to exist was to make RestSharp not have any dependencies.

Issue #1654, opened 2021-11-26
OAuth1Authenticator does not encode ConsumerKey in the authorization header


TrendCare-Alex:
## Read this first!

**Important**
- Please do not use GitHub issues to ask question about using RestSharp.
- Ensure to read the Get help docs page at https://restsharp.dev/get-help/ before opening an issue/ 
- Issues not following our contribution guidelines will be marked as `invalid` and closed in three days.

## Expected Behavior

```csharp
Client.Authenticator = OAuth1Authenticator.ForRequestToken("Consumer+Key/Containing+Reserved+Characters",
                                                            "************",
                                                            OAuthSignatureMethod.HmacSha256);

//  After Client.Authenticator.Authenticate  - Normally performed by Client.Execute(request)

var authHeaderIndex = _request.Parameters.FindIndex(p => p.Name == "Authorization");

_request.Parameters[authHeaderIndex].Value = OAuth Oauth_consumer_key="Consumer%2bKey%2fContaining%2bReserved%2bCharacters",oauth_nonce="i15e67he0d8ycww8",oauth_signature="****************************",oauth_signature_method="HMAC-SHA256",oauth_timestamp="1637878986",oauth_version="1.0"
```

## Actual Behavior

```csharp
Client.Authenticator = OAuth1Authenticator.ForRequestToken("Consumer+Key/Containing+Reserved+Characters",
                                                            "ConsumerSecret",
                                                            OAuthSignatureMethod.HmacSha256);

//  After Client.Authenticator.Authenticate  - Normally performed by Client.Execute(request)

var authHeaderIndex = _request.Parameters.FindIndex(p => p.Name == "Authorization");

_request.Parameters[authHeaderIndex].Value = OAuth Oauth_consumer_key="Consumer+Key/Containing+Reserved+Characters",oauth_nonce="i15e67he0d8ycww8",oauth_signature="OAuthSignature",oauth_signature_method="HMAC-SHA256",oauth_timestamp="1637878986",oauth_version="1.0"
```

## Steps to Reproduce the Problem

  1.  Create an Authenticator on the RestClient Authenticator with a Consumer Key containing reserved characters.
  2.  Invoke the Authenticate method on the Authenticator
  3.  Check the value of the Authorization Header produced by the Authenticate method.

I have been able to get around this problem by doing the following:

```csharp
Client.Authenticator = OAuth1Authenticator.ForRequestToken("Consumer+Key/Containing+Reserved+Characters",
[...]

PR #1671, opened 2021-12-20
Bug/url encode consumer key issue#1654

KennethArn:
## Description

To fix issue #1654. Please note that that this PR add urlencoding to consumer key and in affect removes the generic urldecode. 
Could be a possible breaking change as parameters are not urldecoded by default anymore. I did not find any code that suggest that the parameter could be encoded. 

## Purpose
This pull request is a:

- [] Bugfix (non-breaking change which fixes an issue)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist

### DIFF
--- a/src/RestSharp/Authenticators/OAuth/OAuth1Authenticator.cs
+++ b/src/RestSharp/Authenticators/OAuth/OAuth1Authenticator.cs
@@ -251,7 +251,7 @@ public class OAuth1Authenticator : IAuthenticator {
             => new[] { new Parameter("Authorization", GetAuthorizationHeader(), ParameterType.HttpHeader) };
 
         IEnumerable<Parameter> CreateUrlParameters()
-            => oauth.Parameters.Select(p => new Parameter(p.Name, HttpUtility.UrlDecode(p.Value), ParameterType.GetOrPost));
+            => oauth.Parameters.Select(p => new Parameter(p.Name, p.Value, ParameterType.GetOrPost));
 
         string GetAuthorizationHeader() {
             var oathParameters =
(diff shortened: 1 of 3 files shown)


alexeyzimarev:
Thanks for the PR. But I am not sure what you are fixing here.

The original issue has a problem with headers, but your PR does the query string encoding.

The query string is encoded by default. The authenticator adds all the `oauth_` parameters to the request parameters list. Yes, they aren't encoded if you look inside the request before it actually calls the endpoint. But it should not be tested like this. When the client does the call, it builds the final URL, including the query, and the query parameters get encoded by default.

```csharp
    string? ApplyQueryStringParamsValuesToUri(string? mergedUri, RestRequest request) {
        var parameters = GetQueryStringParameters(request).ToList();
        parameters.AddRange(GetDefaultQueryStringParameters(request));

        if (!parameters.Any()) return mergedUri;

        var separator = mergedUri != null && mergedUri.Contains("?") ? "&" : "?";

        return Concat(mergedUri, separator, EncodeParameters(parameters, Options.Encoding));
    }
```

alexeyzimarev:
Basically, my question is - do you observe any issue with query string `oauth_` parameter encoding, with some specific vendor? Maybe I can run it for that vendor to see if it works or not.

KennethArn:
> Basically, my question is - do you observe any issue with query string `oauth_` parameter encoding, with some specific vendor? Maybe I can run it for that vendor to see if it works or not.

Hi, No I was trying to fix #1654. I reproduced the issue in the test added. But I've have been misunderstanding the issue. Sorry

alexeyzimarev:
No worries, I thought you had an issue yourself.

The problem with your test is that it expects the parameter value to be encoded before the request is made. However, query parameters are encoded (if necessary) _when_ making a request, so all parameter values are not encoded if you just look at the request parameters.

PR #1709, opened 2022-01-18
Support json with arrays on DetectContentType.

MarcelWilke42:
## Description

DetectContentType now supports JSON arrays

## Purpose
This pull request is a:

## Checklist


alexeyzimarev (approved):
Nice catch! Definitely, JSON can start with an object or an array. I might replace it later with a compiled regular expression, but it will work as you suggested.

PR #1746, opened 2022-02-10
Upgrade System.Text.Json to 6.0.*

GieltjE:
## Description

Lots of usefull new things in 6.0.*

## Purpose
This pull request is a:


alexeyzimarev:
Hey, could you please explain why RestSharp needs to get an upgraded version? You can use any version higher or equal than v5 in your project if you want. Setting the lowest acceptable version to 6 will effectively set a dependency constraint for everyone. If you want to use v6, you just need to install it to your project directly, that's all.

Issue #1765, opened 2022-02-24
RestClient constructor: baseUrl property


cosmoscowboy:
In the previous version 106.11.7, the RestClient had a property BaseUrl, but this has been removed since version 107.3.0. Is it possible to add this back to RestClient as a '**_get_**' property?

 I use exports from Postman to send a HttpRequest via RestSharp. The request from Postman contains an absolute Uri. Before sending a RestRequest, I check if the RestClient defines a BaseUrl, and if so, I remove it from the postman request, so that the relative URL is sent in the constructor of the RestRequest. This is no longer possible. I am now forced to use the constructor containing RestClientOptions, as this contains the BaseUrl. 

It seems that all of your documentation creates a RestClient with a base URL, so that the relative URL is sent in the RestRequest. This is logical, as it matches the idea of creating the HttpClient as a singleton, and just sending RestRequest's each time.

cosmoscowboy:
In my previous post I wrote:

`This is no longer possible. I am now forced to use the constructor containing RestClientOptions, as this contains the BaseUrl.`

However, the RestClient doesn't contain a property to the RestClientOptions either, so I am still unable to check what the 'BaseUrl' of the RestClient is!

alexeyzimarev:
I am unsure why do you need to check if the base URL is set or not. It's your client instance, don't you know how it is instantiated?

cosmoscowboy:
The startup file of the .NET core application is creating an instance of the RestClient, so yes, technically another library that uses that client to send a request (that is imported from Postman), doesn't know if the RestClient has a base URL or not. When I read in the URL of the Postman request, it will always contain a host.

Question: assuming the RestClient is instantiated with a BaseUrl, and I create a RestRequest with a resource that is an absolute URL (UriKind.Absolute), is the client smart enough to remove the host, and send on a relative URL? If that is the case, then I would need to do that checking myself.

alexeyzimarev:
The issue you might face is each `RestClient` instance may be configured differently. By reusing an instance you would encounter unexpected behaviour. Say, you have an instance with an authenticator for Twitter, but you call Instagram, what then? Each particular use case needs an instance of `RestClient` that is controlled by that use case. That's why I keep advocating to use specific client API classes, wrapping their own `RestClient` instances, and not exposing `RestClient` out. https://restsharp.dev/usage.html#recommended-usage

cosmoscowboy:
[...]

PR #1782, opened 2022-03-12
Proposed fix for #1765

alexeyzimarev:
* Make `RestClient.Options` public, read-only
* Add a `RestClient` constructor with `HttpClient` only, which won't be changed in any way by RestSharp


IanElsinga:
Thanks, Alexey, I reviewed your changes.  The public read-only Options change looks perfect, and thank you for that.  
I did not have time, however, to assimilate the whole discussion around the other issue (HttpClient) so I cannot really comment on that at this point.

PR #1785, opened 2022-03-13
BaseUrl needs to be internal so it cannot be changed after the client is constructed by the user.

kendallb:
## Description

https://github.com/restsharp/RestSharp/issues/1765


kendallb:
Deleting this and creating a new PR

kendallb:
Move to https://github.com/restsharp/RestSharp/pull/1806

PR #1788, opened 2022-03-14
Remove init setters on RestRequest

kendallb:
These are written to C# 9.0 spec, which means they are init only setter properties. This means any code attempting to consume RestSharp and use these features needs to update their toolchain to support C# 9.0 or the code will not compile. We still have a lot of developers using Visual Studio 2019, which does not support C# 9.0. While we are migrating to VS2022, it's a process to get everyone moved over.

Please change these back to regular setters so they can be used with earlier versions of Visual Studio.

### DIFF
--- a/src/RestSharp/Request/RestRequest.cs
+++ b/src/RestSharp/Request/RestRequest.cs
@@ -1,11 +1,11 @@
 ﻿//  Copyright © 2009-2021 John Sheehan, Andrew Young, Alexey Zimarev and RestSharp community
-// 
+//
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
-// 
+//
 // http://www.apache.org/licenses/LICENSE-2.0
-// 
+//
 // Unless required by applicable law or agreed to in writing, software
 // distributed under the License is distributed on an "AS IS" BASIS,
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
@@ -21,8 +21,8 @@ namespace RestSharp;
 /// Container for data used to make requests
 /// </summary>
 public class RestRequest {
-    readonly Func<HttpResponseMessage, RestResponse>? _advancedResponseHandler;
-    readonly Func<Stream, Stream?>?                   _responseWriter;
+    Func<HttpResponseMessage, RestResponse>? _advancedResponseHandler;
+    Func<Stream, Stream?>?                   _responseWriter;
 
     /// <summary>
     /// Default constructor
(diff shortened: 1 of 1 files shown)


alexeyzimarev:
I think it's a great initiative, I thought about the same.

I only have one question: does it make sense to limit people to set the property once? There's no harm in re-writing the value later in code, as the request lifecycle is limited by one request only, so there's no issue with thread safety, etc.

I also have doubts about those properties in general, as things like an advanced writer aren't really related to the request, they are related to how the response is handled. I'd very much prefer to move those properties to some sort of request/response handling options (similar to JSON serializer options). Issues with that:
- Breaking API (maybe not a big issue, those props aren't mainstream)
- **A lot** of changes in all the extensions, as the options need to be passed over everywhere

kendallb:
> > stream the results of a file download directly into another stream to avoid it allocating a huge array in memory so we can stream large files for downloading
> 
> That's what `DownloadStreamAsync` does better, I believe. The issue with those response writers is that they are synchronous for compatibility.

Ahh, quite likely. I am not familiar with that function. Are there examples of how you would use that? It does bother me that the download code I use now is an async function, but internally it uses ResponseWriter which is not async so blocks. I would much prefer to change it to be a proper async download because the current code I use rendersr to an 8K buffer to pass from stream to stream (blech).

kendallb:
Ahh, I see it replaces Execute. I am going to implement using that. Maybe the response writer callbacks should be marked as deprecated in favor of using DownloadStreamAsync?

alexeyzimarev:
Yeah, I think it would be much cleaner to use client functions to do that kind of work instead of placing these functions to the request, as they have nothing to do with requests.

angelthebault:
Hello,
If I may participate to the discussion. I faced recently a similar situation.
I use DownloadStreamAsync and it works like a charm but I would also need to inspect the response Headers. This is why I was looking at AdvancedResponseWriter but the init only setter property was making my life more difficult.

Do you have an idea of how this use case should be managed ?

Thanks a lot for this wonderful library !

kendallb:
> Hello, If I may participate to the discussion. I faced recently a similar situation. I use DownloadStreamAsync and it works like a charm but I would also need to inspect the response Headers. This is why I was looking at AdvancedResponseWriter but the init only setter property was making my life more difficult.
> 
> Do you have an idea of how this use case should be managed ?

That is a really good question and is the reason for AdvancedResponseWriter. However you can do the same if you set up the RestRequest.OnAfterRequest callback which will get called after the request has been executed, but before the stream is downloads the same as it would for any other request.

alexeyzimarev:
It should be possible to make the response raw bytes content to be lazy, and also add the stream to the response. Then, if you want to inspect the status code, you'd do that, then read the stream. If you ask for bytes, the property would read the stream on its own.

Also, maybe it's a good idea to have a response base, which is content-less. Then, create two types of response - the one we use today, and the one, which has a stream.

kendallb:
So when the request is executed, to get the headers out the entire response needs to be read?

alexeyzimarev:
The advantage would be the ability to inspect the response _before_ downloading the response content.

kendallb:
Right, makes sense. If you just need to inspect the headers and do something, DownloadStreamAsync would work fine with the RestRequest.OnAfterRequest callback, but if you wanted to not download the stream you could not. So yeah, makes sense to me.

kendallb:
Moved to https://github.com/restsharp/RestSharp/pull/1806

PR #1795, opened 2022-03-16
Fix thread safety issues with RestClient when used as a single instance for Authentication and Cookies

kendallb:
## Description

Implements the following changes:
- BaseUrl needs to be internal so it cannot be changed after the client is constructed by the user.
- ResponseWriter and AdvancedResponseWriter are back to regular property setters (no longer init)
- Moved handling of Cookies out of HttpClient and into RestSharp, so they will not cross pollinate requests.
- Move Authenticator from the client into the request, since it's request specific.
- Add back all non-sync functions to RestClient and extensions

## Purpose
Fix issues referenced about thread safety with single instance and improve the API to help developers avoid causing those problems.

## Checklist

### DIFF
--- a/src/RestSharp/Authenticators/JwtAuthenticator.cs
+++ b/src/RestSharp/Authenticators/JwtAuthenticator.cs
@@ -21,13 +21,6 @@ namespace RestSharp.Authenticators;
 public class JwtAuthenticator : AuthenticatorBase {
     public JwtAuthenticator(string accessToken) : base(GetToken(accessToken)) { }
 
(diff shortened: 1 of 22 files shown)


kendallb:
Gonna close this out out and I can cherry pick bits into separate PR's for you and you can merge them. I will rebase this working tree as they get accepted.

bamo-volue:
I see that the extraction of the property Authenticator, within the RestClient, still hasn't been merged to master.
Is it planned to be cherrypicked at all?

PR #1800, opened 2022-03-18
Added back all missing Async API functions

kendallb:
Added back all missing Async API functions (except for json streaming via async not sure how to convert that one).

### DIFF
--- a/src/RestSharp/RestClient.Async.cs
+++ b/src/RestSharp/RestClient.Async.cs
@@ -12,9 +12,17 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+using RestSharp.Extensions;
+
 namespace RestSharp;
 
 public partial class RestClient {
+    /// <summary>
+    /// Executes the request synchronously, authenticating if needed
+    /// </summary>
+    /// <param name="request">Request to be executed</param>
+    public RestResponse Execute(RestRequest request) => AsyncHelpers.RunSync(() => ExecuteAsync(request));
+
     /// <summary>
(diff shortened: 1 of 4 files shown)


src/RestSharp/RestClient.Async.cs
```diff
@@ -12,9 +12,17 @@
...
+using RestSharp.Extensions;
+
 namespace RestSharp;
 
 public partial class RestClient {
+    /// <summary>
+    /// Executes the request synchronously, authenticating if needed
+    /// </summary>
+    /// <param name="request">Request to be executed</param>
+    public RestResponse Execute(RestRequest request) => AsyncHelpers.RunSync(() => ExecuteAsync(request));
```
alexeyzimarev, line 24: Do we need it here, or can it be an extension?

kendallb: Seems simple enough to put it there, but it could be moved to RestClientExtensions I suppose? What would you prefer?

alexeyzimarev: The question is if it should be a part of the _potentially_ returned `IRestClient` interface. Do you think it makes sense to keep non-async methods as the core functionality?

kendallb: Yes, I think they should be part of the core functionality and if there was an IRestClient interface, it should include the non-async functions as well.

kendallb: There could be two interfaces, one for sync and one for non to keep it clean, but IRestClient should inherit both as many people will use both sync and async in their code.

PR #1801, opened 2022-03-18
Move Cookie handling out of HttpClient so we do not cross pollinate requests

kendallb:
## Description

Implements the following:

- Make RestClient.Options internal again and expose the BaseUrl and BaseHost as read only getters on the client.
- Move handling of Cookies out of HttpClient and into RestSharp, so they will not cross pollinate requests.

## Purpose
This pull request is a:

although the breakage is only against stuff that was not released other than in alpha anyway (making Options internal again).

## Checklist

### DIFF
--- a/src/RestSharp/Request/RestRequestExtensions.cs
+++ b/src/RestSharp/Request/RestRequestExtensions.cs
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+using System.Net;
 using System.Text.RegularExpressions;
 using RestSharp.Extensions;
 using RestSharp.Serializers;
@@ -384,6 +385,21 @@ public static class RestRequestExtensions {
             request.AddParameter(name, value);
         }
 
+        return request;
+    }
+
+    /// <summary>
+    /// Adds cookie to the <seealso cref="HttpClient"/> cookie container.
+    /// </summary>
+    /// <param name="request">RestRequest to add the cookies to</param>
+    /// <param name="name">Cookie name</param>
+    /// <param name="value">Cookie value</param>
+    /// <param name="path">Cookie path</param>
+    /// <param name="domain">Cookie domain, must not be an empty string</param>
+    /// <returns></returns>
+    public static RestRequest AddCookie(this RestRequest request, string name, string value, string path, string domain) {
+        request.CookieContainer ??= new CookieContainer();
+        request.CookieContainer.Add(new Cookie(name, value, path, domain));
         return request;
     }
 
--- a/src/RestSharp/RestClient.cs
+++ b/src/RestSharp/RestClient.cs
@@ -27,8 +27,6 @@ namespace RestSharp;
 /// Client to translate RestRequests into Http requests and process response result
 /// </summary>
 public partial class RestClient : IDisposable {
-    public CookieContainer CookieContainer { get; }
-
     /// <summary>
     /// Content types that will be sent in the Accept header. The list is populated from the known serializers.
     /// If you need to send something else by default, set this property to a different value.
@@ -45,13 +43,12 @@ public partial class RestClient : IDisposable {
 
     HttpClient HttpClient { get; }
 
-    public RestClientOptions Options { get; }
+    internal RestClientOptions Options { get; }
 
     public RestClient(RestClientOptions options, Action<HttpRequestHeaders>? configureDefaultHeaders = null) {
         UseDefaultSerializers();
 
         Options            = options;
-        CookieContainer    = Options.CookieContainer ?? new CookieContainer();
         _disposeHttpClient = true;
 
         var handler = new HttpClientHandler();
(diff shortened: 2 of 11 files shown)


src/RestSharp/Request/RestRequestExtensions.cs
```diff
@@ -387,6 +388,23 @@ public static RestRequest AddXmlBody<T>(this RestRequest request, T obj, string
...
+    /// Adds cookie to the <seealso cref="HttpClient"/> cookie container.
+    /// </summary>
+    /// <param name="request">RestRequest to add the cookies to</param>
+    /// <param name="name">Cookie name</param>
+    /// <param name="value">Cookie value</param>
+    /// <param name="path">Cookie path</param>
+    /// <param name="domain">Cookie domain, must not be an empty string</param>
+    /// <returns></returns>
+    public static RestRequest AddCookie(this RestRequest request, string name, string value, string path, string domain) {
+        if (request.CookieContainer == null) {
```
alexeyzimarev, line 401: We can create a cookie container on the first call of `AddCookie`

src/RestSharp/RestClient.cs
```diff
@@ -123,15 +123,25 @@ public partial class RestClient : IDisposable {
     /// <param name="disposeHandler">Dispose the handler when disposing RestClient, true by default</param>
     public RestClient(HttpMessageHandler handler, bool disposeHandler = true) : this(new HttpClient(handler, disposeHandler), true) { }
 
+    /// <summary>
+    /// Returns the currently configured BaseUrl for this RestClient instance
+    /// </summary>
+    public Uri BaseUrl => Options.BaseUrl!;
```
alexeyzimarev, line 129: If the `Options.BaseUrl` is nullable, this property should also be nullable.

src/RestSharp/RestClient.cs
```diff
@@ -123,15 +123,25 @@ public partial class RestClient : IDisposable {
...
+
+    /// <summary>
+    /// Returns the currently configured BaseHost for this RestClient instance
+    /// </summary>
+    public string BaseHost => Options.BaseHost!;
```
alexeyzimarev, line 134: Same here

src/RestSharp/Request/RestRequestExtensions.cs
kendallb: Yeah good point. I will change it.

src/RestSharp/RestClient.cs
kendallb: Yep will fix this also.

kendallb:
Rebased against dev. Still need to decide how to handle redirects. For now I have simply disabled redirects in the bits of code that caused problems and process it outselves (we didn't actually need to follow it anyway). But it's not a generic solution.

Honestly it might be good enough and to document to the user that cookies won't follow redirects. 99% of the time REST services do not use cookies at all, so this is really just to make sure the functionality actually works correctly for those who have used it.

bednar:
Hi @kendallb,

For our purpose it would be convenient to give us the option to redirect the cookies to another domain. Our users uses cookies to  authentication. But at the same time they have a proxy set up that performs the redirection to cloud service.

- https://github.com/influxdata/influxdb-client-csharp/pull/302

Regards

alexeyzimarev:
The Influx case will get worse if cookies aren't held in the client. I believe having a cookie container for the client should not be removed entirely, but allowed to be switched off.

kendallb:
I completely disagree here and I am not sure you fully appreciate the problem? The cookie container is being shared across all requests, which is bad, really bad. Even Microsoft says in their documentation to not use the cookie container at all. The solution I built allows the cookies to be request specific which is what they need to be.

If the users wants them to persist between their own requests, they will install their own container. That is how we did it previously and how we still do it now with our own rest sharp client library.

alexeyzimarev:
Could you point to the documentation where they say "don't use cookie container"? Why does it exist? I think you are referring to the HttpClientFactory docs where they say that the factory cannot be used if you need to use a custom cookie container.

Here's the case. I have a client for an external API. It uses authorisation cookies. After the initial authorisation, the cookies must be kept in the container and sent with any consequent request.

The case when cookies _must not_ be shared between requests if the client is used for multiple tenants (for example).

These are two different use cases.

alexeyzimarev:
I checked the CookigContainer code and it's thread-safe, they use locks in the `Add` method.

Don't get me wrong, I don't argue with the need to handle cookies per request. The only thing I am saying is that we should consider both use cases, and clearly separate them. I really like your PR where cookies are added per request as headers, I just don't want to remove the custom cookie container from the client entirely in case people really need to use it.

kendallb:
Well the pros to leaving it in the RestClient itself is that folks who might want a simple way to persist cookies can have that, but the big con IMHO is that users of the library see that and then make the fatal mistake of thinking that either a) they should be using RestClient as instance per request like before (as Easy Post did and still has not fixed and ignored my PR!) or b) they end up using it without realizing the ramifications of the shared cookie container across requests.

b) is my biggest concern because it's a glaring security hold. a) is a concern because the more 'code smells' people run across trying to move to v107 and later the more they finally are forced to realize that they are doing it wrong and need to change their code. 

It would certainly be possible to get the container from either the client or the request, but I think it's too dangerous to leave it in the client? Folks will ask why it was moved, and once they understand why I think they would be fine doing things differently.

alexeyzimarev:
Yeah, I agree. I think your change to delete the new CookieContainer instance creation from the client is good. If the default container isn't set, the chance of hitting this issue is basically the same as it's with HttpClient.

kendallb:
Also 99% of folks using RestSharp for REST services generally do not use cookies anyway so it won't affect that many folks, and for those who do get affected it probably matters to get it right.

PR #1805, opened 2022-03-18
Change RestBaseResponse property bag bag to setters from init

kendallb:
## Description

These need to not be init only, because it can be quite useful to return mocked responses and it these are all init only, we cannot adjust them after creating a new instance.

I guess you could argue we should mock the class instead, so I could try that also, but is there any reason not to allow these properties to be setters so its easier for faking responses from lower level services?

## Purpose
This pull request is a:

### DIFF
--- a/src/RestSharp/Response/RestResponse.cs
+++ b/src/RestSharp/Response/RestResponse.cs
@@ -30,7 +30,7 @@ public class RestResponse<T> : RestResponse {
     /// <summary>
     /// Deserialized entity data
     /// </summary>
-    public T? Data { get; internal set; }
+    public T? Data { get; set; }
 
     public static RestResponse<T> FromResponse(RestResponse response)
         => new() {
--- a/src/RestSharp/Response/RestResponseBase.cs
+++ b/src/RestSharp/Response/RestResponseBase.cs
@@ -38,67 +38,67 @@ public abstract class RestResponseBase {
     /// <summary>
     /// MIME content type of response
     /// </summary>
-    public string? ContentType { get; init; }
+    public string? ContentType { get; set; }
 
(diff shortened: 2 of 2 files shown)


kendallb:
I cannot run that Twitter test, as I don't have keys. How do I set it up to run that test? 

I suspect it's failing because Twitter is down or something?

kendallb:
Oh, its not twitter, that is skipped. I am not sure what is failing? It passes for me?

Passed!  - Failed:     0, Passed:    81, Skipped:     7, Total:    88, Duration: 1 m 45 s - RestSharp.Tests.Integrated.dll (net6.0)
Error: Process completed with exit code 1.

kendallb:
I did get a transient failure on Task_Handles_Non_Existent_Domain, but I ran it again and it worked. So I wonder if it's the same issue.

kendallb:
> I guess you could argue we should mock the class instead, so I could try that also, but is there any reason not to allow these properties to be setters so its easier for faking responses from lower level services?

Bah, that does not work as you cannot mock anything on a real class unless the values are virtual so the only way to mock this would be via an interface (which is gone), or by creating a mock real response property bag and stuffing values into it.

alexeyzimarev:
> These need to not be init only, because it can be quite useful to return mocked responses and it these are all init only, we cannot adjust them after creating a new instance.

The way to work with records  is to use something:

```csharp
var newResp = response with { Content = "blah" };
```

The reason for those props to be init-only is to enforce immutability. The response should not be changed after it's created. I only see an issue with `init` that it doesn't work with older compiler versions, but that would be fixed by making them `internal set`. Could you give an example of the mocked response where you need to manipulate its content?

PR #1806, opened 2022-03-18
Move BaseHost, CachePolicy and UserAgent out of the RestClient and into the request and more

kendallb:
## Description

Move BaseHost, CachePolicy and UserAgent out of the RestClient and into the request as those are request specific things and should be something you can change on a per request basis with a singleton client.

Made ResponseWriter and AdvancedResponseWriter both setter properties again so they can be changed after the request has been created, and will also work with compilers prior to the C# 9.0 spec.

## Purpose
This pull request is a:

## Checklist


hcasado (approved):

kendallb:
Rebased against dev

kendallb:
Yeah that makes sense to allow it as an override. If the client is shared it really should not be set in the client but in some cases it might make sense.

kendallb:
Ok I have updated this PR to have it use the default from the client. I am still unclear on why we do this exactly?

```
if (HttpClient.DefaultRequestHeaders.UserAgent.All(x => x.Product?.Name != "RestSharp")) {
    // add headers           
}
```

What is the reason to check the base client for default headers containing RestSharp, and to NOT include the ones for the request in that case? I think 99.9% of the time the HttpClient is going to have no default headers anyway, but I suppose someone could add them. But if they added their own, why would they make them RestSharp headers? We add our own to identify our own requests, not generic RestSharp ones. 

So it seems to me that check should simply be dropped, and we would assume if someone wanted to add their own user agent header at the HttpClient level they would turn it off at the RestSharp level?

alexeyzimarev:
The reason for the constructor that accepts an external `HttpClient` instance is for people to pass a pre-configured instance. I don't want to make assumptions about how that client is configured. 

However, I would agree that if the user agent header is set for the client, we should not be checking for `RestSharp` string, and just not try to set anything at all. Btw, it's not my code, I think someone made a PR. The previous version kept adding the user-agent header when the `HttpClient` instance was reused.

kendallb:
Right, makes sense in the context of where it used to be on the HttpClient, but not in the context of getting it for the request so I removed that.

alexeyzimarev:
I got to think that both ways need to be there, and request properties would override the client options.

I still struggle to understand how to make the client options immutable as people keep overriding them between requests, expecting the client behaviour to change. I am considering to hide the `Options` property as it's easily abused. I don't remember the motivation to make it public, but it was internal by default, and there were some complaints about it.

It might get along with the coming factory. Not sure about that one too, but Flurl simply decided to cache client instances using the base URL to separate them. Maybe we can do the same. Unless I find a way to accommodate the `IHttpClientFactory`, but then I'd need to support named configs for `RestClient`.

kendallb:
Well in our case we need to change options for our clients so I create a cache key based on the options which includes the base URL. That seems to be working well for us so far but I also do not recall why Client.Options needs to be public. In our case we pass in the options class to our client factory, and internally create the key from all the options that we need to vary on, and then either use a cache client or create a new one. We never look at client.Options again after that, so I think it would be fine to hide it so folks know it is immutable?

The one thing we did just get bitten by (unrelated to this) and something I have not yet added a PR for but have published it in our own version, is accidentally disposing of a cached client. This caused an error deep inside the handler so I changed the code to throw a disposed exception much earlier up as it took a while to figure out what was failing as the request processed just fine, but we had an internal exception in the rest response we forgot to look for. In cases like that I would prefer RestSharp throw a real exception as something is terribly wrong at that point.

https://github.com/kendallb/RestSharp/commit/27c47663a0ea9bf10206125d49fdf00d1a83d10f

The other change I also made in there was to throw an exception on non-success status codes for DownloadStreamAsync because it will just download the contents of a 404 page if you get a 404 which is not what you would really expect. So that now throws a WebException that mimics pretty much what the old WebRequest code used to do for download API calls.

alexeyzimarev:
Yeah, checking if the client is disposed makes total sense.

> Well in our case we need to change options for our clients so I create a cache key based on the options which includes the base URL.

I believe that's what I am thinking about. What would you need to change? The question is where the cache should reside. I'd rather have it in the factory, but I know that people keep instantiating the client per request. Maybe the factory can have a default static instance, so I can select the cached `HttpClient` instance from there? Maybe add some function to disable the cache, but it would be an opt-out.

kendallb:
The cache exists in the factory in our case. This is how ours is written and it's a static class so the user just calls it to get a static instance any time it's needed (no DI):

```C#
/*
 * Copyright (C) 2004 and later AMain.com, Inc.
 * All Rights Reserved
 */

using System;
using System.Collections.Generic;
using System.Net;
using System.Text;
using Newtonsoft.Json;
using Newtonsoft.Json.Serialization;
using RestSharp;
using RestSharp.Serializers;
using RestSharp.Serializers.Json;
using RestSharp.Serializers.NewtonsoftJson;
using RestSharp.Serializers.Xml;

namespace AMain.RestClientHelper
{
    public static class RestClientFactory
    {
        /// <summary>
        /// Dictionary of all static client instances based on the client API base URL
        /// </summary>
        private static readonly Dictionary<string, RestClient> _clientCache = new();

        /// <summary>
        /// Get a client bound to a specific URL that uses only JSON for serialization
        /// </summary>
        /// <param name="cachePrefix">Cache prefix to use for the client cache</param>
        /// <param name="baseUrl">The client base URL</param>
        /// <param name="jsonNaming">JSON naming convention to use</param>
        /// <param name="timezone">Optional timezone for this client for timezone conversions</param>
        /// <param name="disableSslChecks">True to disable SSL certificate checks</param>
        /// <returns>RestClient instance to use</returns>
        public static RestClient GetJsonClient(
            string cachePrefix,
            string baseUrl,
            JsonNaming jsonNaming,
            TimeZoneInfo timezone = null,
            bool disableSslChecks = false)
        {
            // Change to use Json.net for with custom serialization to match how the old SimpleJson serializer worked
            // as much as possible, but with different potential naming conventions
            IRestSerializer JsonSerializerFactory()
            {
                // Make sure we enable inclusion of null values (thanks UPS!) and also make sure we can handle
                // converting single elements to a list (thanks again UPS!).
                var settings = new JsonSerializerSettings {
                    NullValueHandling = NullValueHandling.Include,
                    Converters = {
                        new SingleOrArrayListConverter(),
                    },
                };

                // If we have an optional timezone, make sure it gets added into the converter chain
                if (timezone != null) {
                    settings.Converters.Add(new TimeZoneDateTimeConverter(timezone) {
                        WriteAsUniversal = true,
                        ReadAsLocalTime = true,
                    });
                }

                // Optionally change the naming convention. If we do not set anything it will just use the original C# property name
                // as the name used in the resulting JSON.
                switch (jsonNaming) {
                    case JsonNaming.Lowercase:
                        settings.ContractResolver = new DefaultContractResolver {
                            NamingStrategy = new LowerCaseNamingStrategy(),
                        };
                        break;
                    case JsonNaming.CamelCase:
                        settings.ContractResolver = new DefaultContractResolver {
                            NamingStrategy = new CamelCaseNamingStrategy(),
                        };
                        break;
                    case JsonNaming.SnakeCase:
                        settings.ContractResolver = new DefaultContractResolver {
                            NamingStrategy = new SnakeCaseNamingStrategy(),
                        };
                        break;
                }
                return new JsonNetSerializer(settings);
            }

            // If we have an optional timezone for this client, we MUST adjust the cache prefix as we need a different client for each timezone!
            var sb = new StringBuilder(cachePrefix);
            if (timezone != null) {
                sb.Append(':').Append(timezone.BaseUtcOffset);
            }
            var options = string.IsNullOrEmpty(baseUrl) ? new RestClientOptions() : new RestClientOptions(baseUrl);
            return InternalGetClient(sb, options, disableSslChecks, JsonSerializerFactory);
        }

        /// <summary>
        /// Get a client bound to a specific URL that uses only XML for serialization
        /// </summary>
        /// <param name="cachePrefix">Cache prefix to use for the client cache</param>
        /// <param name="baseUrl">The client base URL</param>
        /// <param name="disableSslChecks">True to disable SSL certificate checks</param>
        /// <returns>RestClient instance to use</returns>
        public static RestClient GetXmlClient(
            string cachePrefix,
            string baseUrl,
            bool disableSslChecks = false)
        {
            return InternalGetClient(new StringBuilder(cachePrefix), new RestClientOptions(baseUrl), disableSslChecks, () => new XmlRestSerializer());
        }

        /// <summary>
        /// Get a client bound to a specific set of client options. We store static instances of RestClient for each
        /// base URL and related options as we do not want to have an instance per request, and we can cache them statically
        /// and reuse them over and over as they are thread safe provided we fully configure the RestClient in here.
        /// </summary>
        /// <param name="cachePrefix">Cache prefix to use for the client cache</param>
        /// <param name="options">The client options for this client</param>
        /// <param name="disableSslChecks">True to disable SSL certificate checks</param>
        /// <param name="serializerFactory">Optional custom serializer factory to use for requests and responses</param>
        /// <returns>RestClient instance to use</returns>
        public static RestClient GetClient(
            string cachePrefix,
            RestClientOptions options,
            bool disableSslChecks = false,
            Func<IRestSerializer> serializerFactory = null)
        {
            return InternalGetClient(new StringBuilder(cachePrefix), options, disableSslChecks, serializerFactory);
        }

        /// <summary>
        /// Get a basic client bound to a the default options that we can re-use over and over.
        /// </summary>
        /// <param name="disableSslChecks">True to disable SSL certificate checks</param>
        /// <returns>RestClient instance to use</returns>
        public static RestClient GetBasicClient(
            bool disableSslChecks = false)
        {
            return InternalGetClient(new StringBuilder("BasicClient"), new RestClientOptions(), disableSslChecks, null);
        }

        /// <summary>
        /// Get a client bound to a specific set of client options. We store static instances of RestClient for each
        /// base URL and related options as we do not want to have an instance per request, and we can cache them statically
        /// and reuse them over and over as they are thread safe provided we fully configure the RestClient in here.
        /// </summary>
        /// <param name="cachePrefix">Cache prefix to use for the client cache</param>
        /// <param name="options">The client options for this client</param>
        /// <param name="disableSslChecks">True to disable SSL certificate checks</param>
        /// <param name="serializerFactory">Optional custom serializer factory to use for requests and responses</param>
        /// <returns>RestClient instance to use</returns>
        private static RestClient InternalGetClient(
            StringBuilder cachePrefix,
            RestClientOptions options,
            bool disableSslChecks,
            Func<IRestSerializer> serializerFactory)
        {
            // Bail if we see unsupported features
            if (options.ConfigureMessageHandler != null) {
                throw new ArgumentException("RestClientFactory does not support ConfigureMessageHandler!", nameof(RestClientOptions));
            }
            if (options.Credentials != null || options.UseDefaultCredentials) {
                throw new ArgumentException("RestClientFactory does not support Credentials!", nameof(RestClientOptions));
            }
            if (options.ClientCertificates != null) {
                throw new ArgumentException("RestClientFactory does not support ClientCertificates!", nameof(RestClientOptions));
            }
            if (options.RemoteCertificateValidationCallback != null) {
                throw new ArgumentException("RestClientFactory does not support RemoteCertificateValidationCallback!", nameof(RestClientOptions));
            }
            if (options.MaxTimeout > 0) {
                throw new ArgumentException("RestClientFactory does not support MaxTimeout! Please set it at the request level", nameof(RestClientOptions));
            }

            // Set the user agent so we can identify our own requests
            options.UserAgent = "SNAPMerchant/1.0";

            // Determine the string to use for the proxy if passed in
            var proxyString = "";
            if (options.Proxy is WebProxy proxy) {
                if (proxy.Credentials != null) {
                    var c = proxy.Credentials as NetworkCredential;
                    proxyString = $"http://{c.UserName}:{c.Password}@{proxy.Address.Authority}/";
                }
                proxyString = proxy.Address.AbsoluteUri;
            }

            // Build a key to represent the client in the cache
            cachePrefix.Append(':')
                .Append(options.BaseUrl).Append(':')
                .Append(disableSslChecks).Append(':')
                .Append(options.DisableCharset).Append(':')
                .Append(options.AutomaticDecompression).Append(':')
                .Append(options.FollowRedirects).Append(':')
                .Append(options.MaxRedirects).Append(':')
                .Append(proxyString).Append(':')
                .Append(options.Encoding).Append(':')
                .Append(options.PreAuthenticate).Append(':')
                .Append(options.Expect100Continue);
            var key = cachePrefix.ToString();

            // If we have an existing instance of this client, return it outside of the lock
            if (_clientCache.TryGetValue(key, out var client)) {
                return client;
            }

            // Create a new instance, but make sure only one thread can create it
            lock (_clientCache) {
                // Try again to get it, as it might have just been created by another thread
                if (_clientCache.TryGetValue(key, out client)) {
                    return client;
                }

                // Disable the SSL certificate checks if needed
                if (disableSslChecks) {
                    options.RemoteCertificateValidationCallback = (_, _, _, _) => true;
                }

                // Set up the client with an infinite timeout so the request timeout will be the one that always matters
                options.MaxTimeout = int.MaxValue;

                // Create the client and cache it
                client = new RestClient(options);

                // Set up the single serializer we want to use for this client if desired. If this is passed in as null,
                // it will use the defaults which is both the XML and System.Text.Json serializers.
                if (serializerFactory != null) {
                    client.UseOnlySerializer(serializerFactory);
                } else {
                    // We need to reverse the order of the serializers here, as the default with RestSharp is not JSON and then XML
                    // but we have code that assumes that XML will get used first (QBP). So easier to flip the defaults. Setting
                    // the only serializer first to XML clears the list, then we can add in JSON and it will be second.
                    client.UseOnlySerializer(() => new XmlRestSerializer());
                    client.UseSerializer(() => new SystemTextJsonSerializer());
                }

                _clientCache.Add(key, client);
                return client;
            }
        }
    }
}
```

Issue #1816, opened 2022-03-29
Can no longer add JSON manually to a request as a parameter


kendallb:
**Describe the bug**
In RestSharp v106 and earlier, it was possible to add JSON to the request using the following sequence:

```C#
request.AddParameter("application/json", myJsonString, ParameterType.RequestBody);
```

With RestSharp v107 this no longer works. Instead you get a request like this:

```
POST /soa-infra/resources/GWS/ProductAvailability/Availability HTTP/1.1
User-Agent: Mozilla/5.0 (compatible; MSIE 9.0; Windows NT 6.1; Trident/5.0)
Accept: application/xml, text/xml, application/json, text/json, text/x-json, text/javascript
Content-Type: application/json; boundary="030269ab-0029-48b8-b40e-d12d4445826e"
Content-Length: 197
Host: ws.specialized.com
Cache-Control: no-cache
Connection: keep-alive
--030269ab-0029-48b8-b40e-d12d4445826e
Content-Type: text/plain; charset=utf-8
Content-Disposition: form-data; name="application/json"
System.Byte[]
--030269ab-0029-48b8-b40e-d12d4445826e--
```

when it should be like this:

```
POST /soa-infra/resources/GWS/ProductAvailability/Availability HTTP/1.1
User-Agent: Mozilla/5.0 (compatible; MSIE 9.0; Windows NT 6.1; Trident/5.0)
Accept: application/xml, text/xml, application/json, text/json, text/x-json, text/javascript
Content-Type: application/json
Content-Length: 232
Host: ws.specialized.com
Cache-Control: no-cache
Connection: keep-alive
{"CredentialRequest":{"Username":"XXXX", ... [rest of JSON snipped]
```

**To Reproduce**
Try to add JSON manually to the body of the request as above.

**Expected behavior**
It should behave the same as is you had called the following, which is how v107 expects you to do it:

```C#
request.AddBody(myJsonString, "application/json");
```

While you could argue that might be the 'correct' way, the problem is the old libraries fully supported adding the request body that way and the API *still* appears to support it as it has the correct function signature for it. Anyone porting existing code will run into this so it should be fixed.

**Desktop (please complete the following information):**
 - Windows 10, Windows Server 2016
 - .NET Framework
 - Version 4.8

kendallb:
Ok fixed in this PR with unit test that fails without the fix, and passes with it:

https://github.com/restsharp/RestSharp/pull/1817

The problem is this line:

[...]

PR #1817, opened 2022-03-29
Fix for manual JSON body content parameter

kendallb:
## Description
Fix for https://github.com/restsharp/RestSharp/issues/1816

## Purpose
This pull request is a:

## Checklist


src/RestSharp/Parameters/Parameter.cs
```diff
@@ -34,7 +34,7 @@ public static Parameter CreateParameter(string? name, object? value, ParameterTy
             ParameterType.GetOrPost   => new GetOrPostParameter(name!, value?.ToString(), encode),
             ParameterType.UrlSegment  => new UrlSegmentParameter(name!, value?.ToString()!, encode),
             ParameterType.HttpHeader  => new HeaderParameter(name, value?.ToString()),
-            ParameterType.RequestBody => new BodyParameter(name, value!, Serializers.ContentType.Plain),
+            ParameterType.RequestBody => new BodyParameter("", value!, string.IsNullOrEmpty(name) ? Serializers.ContentType.Plain : name!),
```
alexeyzimarev, line 37: That will break things. Unlike RS 106, the latest version allows you to add multiple body parameters. It works with multipart form data. There, parameters must have names, also body parameters.

kendallb: Yeah, not sure I like the fix either but so much code does it that way :(

alexeyzimarev: The question is: do we want to make it compatible or force people to use the right API. In the latter case, we can throw an exception if the parameter name looks like content type, and the exception would say: "it won't work, use `AddStringBody`". I can add an overload for `AddStringBody` that would accept the parameter name.

kendallb: That’s a good idea. For us it took a long time to figure out what was wrong as I had to inspect the outgoing packets to finally work out it was not sent with the right content type. A logical exception explaining what is wrong would save those folks with the same problem a lot of time.

kendallb:
Checked #1869 and it looks like a good solution for me. Deleting my branch.

alexeyzimarev:
Btw, Postman is being updated, they asked me to review the new generator version.

PR #1827, opened 2022-04-04
Remaining patches all in one commit

kendallb:
## Description
Since I have quite a few intertwined PR's left, some of those won't rebase onto each other easily. So here is one that has it all combined if you want to review that and we discuss any tweaks necessary.

Otherwise you can continue to review my other changes, but until these are accepted in some form we still have to keep using our own fork in production.

### DIFF
--- a/src/RestSharp/Request/RestRequest.cs
+++ b/src/RestSharp/Request/RestRequest.cs
@@ -13,6 +13,8 @@
 // limitations under the License.
 
 using System.Net;
+using System.Net.Http.Headers;
+using RestSharp.Authenticators;
 using RestSharp.Extensions;
 // ReSharper disable UnusedAutoPropertyAccessor.Global
 
@@ -22,8 +24,8 @@ namespace RestSharp;
 /// Container for data used to make requests
 /// </summary>
 public class RestRequest {
-    readonly Func<HttpResponseMessage, RestResponse>? _advancedResponseHandler;
-    readonly Func<Stream, Stream?>?                   _responseWriter;
+    Func<HttpResponseMessage, RestResponse>? _advancedResponseHandler;
+    Func<Stream, Stream?>?                   _responseWriter;
 
     /// <summary>
     /// Default constructor
@@ -111,7 +113,11 @@ public class RestRequest {
     public Method Method { get; set; }
 
     /// <summary>
-    /// Custom request timeout
+    /// Sets the timeout in milliseconds for this requests using this client. Note that there is also a timeout
+    /// set on the base client, and the the shorter of the two values is what will end up being used. So if you need long
+    /// timeouts at the request level, you will want to set the value on the client to to a larger value than the maximum
+    /// you need per request, or set the client to infinite. If this value is 0, an infinite timeout is used (basically
+    /// it then times out using whatever was configured at the client level).
     /// </summary>
     public int Timeout { get; set; }
 
@@ -171,11 +177,31 @@ public class RestRequest {
     public HttpCompletionOption CompletionOption { get; set; } = HttpCompletionOption.ResponseContentRead;
 
     /// <summary>
+    /// Explicit Host header value to use in requests independent from the request URI.
+    /// </summary>
+    public string? BaseHost { get; set; }
+
+    /// <summary>
+    /// Sets the user agent string to be used for this requests. Defaults to a the client default if not provided.
+    /// </summary>
+    public string? UserAgent { get; set; }
+
+    /// <summary>
+    /// Sets the cache policy to use for this request
+    /// </summary>
+    public CacheControlHeaderValue? CachePolicy { get; set; }
+
+    /// <summary>
+    /// Authenticator that will be used to populate request with necessary authentication data
+    /// </summary>
+    public IAuthenticator? Authenticator { get; set; }
+
+    /// <summary>
     /// Set this to write response to Stream rather than reading into memory.
     /// </summary>
     public Func<Stream, Stream?>? ResponseWriter {
         get => _responseWriter;
-        init {
+        set {
             if (AdvancedResponseWriter != null)
                 throw new ArgumentException(
                     "AdvancedResponseWriter is not null. Only one response writer can be used."
@@ -190,7 +216,7 @@ public class RestRequest {
     /// </summary>
     public Func<HttpResponseMessage, RestResponse>? AdvancedResponseWriter {
         get => _advancedResponseHandler;
-        init {
+        set {
             if (ResponseWriter != null)
                 throw new ArgumentException("ResponseWriter is not null. Only one response writer can be used.");
 
--- a/src/RestSharp/Request/RestRequestExtensions.cs
+++ b/src/RestSharp/Request/RestRequestExtensions.cs
@@ -14,6 +14,7 @@
 
 using System.Net;
 using System.Text.RegularExpressions;
+using RestSharp.Authenticators;
 using RestSharp.Extensions;
 using RestSharp.Serializers;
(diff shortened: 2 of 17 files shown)


kendallb:
It has been a while since I wrote this and I really need to rebase is against the current develop tree to determine what needs to stay. If you cherry pick over what you will accept now I can go back and rebase it and look at the remaining bits we changed and research why. We made all those changes to fix things we ran into using it in our code, but I am sure the approach can be changed to achieve what we needed and be acceptable upstream.

kendallb:
I just rebased this against develop so I can get a clean merge and will go through and revisit everything. Some stuff was already merged so I cleaned that up in the rebase.

kendallb:
Ok following up on your comments:

> * Adding an authenticator to the request makes sense, but I am not sure the way it's done here is right. I'll give it more thought.

The authenticator must be request based if the client is shared across requests that might need different authentication (ie: different for each user of a web site for instance so different in each request). We have that case so that is what we needed it for. I removed it from the base client because I believe that is the wrong location for the authenticator to reside because it is inherently a request centric thing and not related to the client instance at all.

But it is a breaking change so I would suggest leaving it in the client for now and make the request based version override the client version, and only use the client version if the request one is not present.

But then I would seriously consider removing it like I have done in v109 as a breaking change as I think that is another way to drill home the concept that the client should be reused for multiple requests and not instantiated per request.

> * Removing the option from `StreamJsonAsync` is wrong; streaming won't work without it.

I think this was an artifact of the old code base, as now that I have rebased all that happened to StreamJsonAsync was to add the optional authenticator. The option for the headers was moved around in a prior commit which was already in develop so that bit is clean now.

> * Request-level cookies handling looks good!

Thanks. It's a bit merged in with the rest of the code but I do have a branch with mostly the cookie stuff in it (and I have rebased it) if that helps. That still does have the base URL stuff in it, and moving the options class around and making it private in the client again but that could be moved out that change. LMK if you want me to do that.

> * Cookie container on the client level should stay, as it works fine for machine-2-machine scenarios (as well as the client-level authenticator)

I would argue it is dangerous to let the cookie container remain at the client level because you will have cookies end up crossing over requests for any reused clients. If the client happens to be used in a multi user environment such as on a web site, you would never want cookies from the request for one user to an external service to end up in the request for another user of the same service which is exactly what will happen if the cookies are shared at the client level.

So while it's plausible that it can be used in some scenarios, I would argue it is way too dangerous to allow that to happen and better to break the API and have folks update their code to do cookie handling at the request level.

As I mentioned previously when it comes to cookie handling, 99% of folks using RestSharp for interacting with REST services won't be using cookies anyway, as it's is very unusual for stateless REST services to care about cookies. In our case the only time we are actually using cookies is when we are doing web scraping using RestSharp and need to process user logins, so we execute the login code to get the cookies and then requests after that are properly authenticated. It is certainly vital in that scenario that the cookies are never shared.

> * I am not sure what's the reason for removing `SetBearerToken`, as it was added by a contributor, so they needed it.

Yeah not sure why I did that, so I removed it.

> * Using content type as the parameter name is legacy, and I don't think I want to get it back tbh. I really want to remove as many implicit things as possible, and this is the worst one.

Yes, all that is removed now as you implemented a better solution anyway a while back.

kendallb:
BTW I think we discussed somewhere else about the public Options class in the client. In my code I made it internal from the start and exposed BaseUrl and Host at the client level as properties. That is a breaking change so perhaps something to leave out until v109?

kendallb:
BTW, related to client level cookies, the old client always created a cookie container in the client which is always used for all requests. As mentioned 99% of RestSharp users will not be using cookies, however having a container at the client level means cookies *will* be cross pollinating between requests using the same client instance which is really bad (big security issue). So most of the time there probably are not any cookies that matter, but if there are it could be problematic for the user of the library. 

Recall Microsoft even suggests not using it at all with HttpClient:

https://docs.microsoft.com/en-us/aspnet/core/fundamentals/http-requests?view=aspnetcore-6.0

**Cookies**
The pooled HttpMessageHandler instances results in CookieContainer objects being shared. Unanticipated CookieContainer object sharing often results in incorrect code. For apps that require cookies, consider either:

- Disabling automatic cookie handling
- Avoiding IHttpClientFactory

The biggest annoying issue with the way cookies work in HttpClient is that they do not follow with redirects if redirects are followed within HttpClient itself (something we have run into). I don't have a clean solution for that at the moment without finding a way to process the redirects somehow at the RestSharp level but I just worked around that in our code.

kendallb:
Created a new pull request with just the cookie handling and removed the bits changing the client Options to be internal. I still think we should do that, but perhaps that is better in another pull request?

https://github.com/restsharp/RestSharp/pull/1966

kendallb:
Oh I remember now why I removed SetBearerToken(). Because the authenticator is now per request, you do not need to set the bearer token for the authenticator attached to the client as you can just create a new one. So the reason for needing it at the authenticator level is gone (and I updated the unit test to reflect that). But it can certainly be left there so someone can re-use an authenticator and adjust the bearer token as needed. I will adjust the test to show both cases.

alexeyzimarev:
Just remember that for M2M integration, everything is reused. Say I want to call a service on Cloud Run. I need to get the application service account, get the id token, and the authenticator would generate an access token. The access token will be used for all the requests until it expires. The scenario of forcing the authenticator to handle 401 is something I am considering, as it would be very useful. This `SetToken` thing is how people do it now (manually).

kendallb:
> Just remember that for M2M integration, everything is reused. Say I want to call a service on Cloud Run. I need to get the application service account, get the id token, and the authenticator would generate an access token. The access token will be used for all the requests until it expires. The scenario of forcing the authenticator to handle 401 is something I am considering, as it would be very useful. This `SetToken` thing is how people do it now (manually).

Fair enough, so you can just use a shared authenticator or create a new one using the correct token for each request. Since the authenticator is now set at the request level, the need to override the token in an existing authenticator is gone as that was really only a necessary requirement when the authenticator was set on the client and could not be changed once the client was constructed.

alexeyzimarev:
> the old client always created a cookie container in the client which is always used for all requests

Yeah, it's a design mistake, makes no sense to do it.

alexeyzimarev:
> BTW I think we discussed somewhere else about the public Options class in the client. In my code I made it internal from the start and exposed BaseUrl and Host at the client level as properties. That is a breaking change so perhaps something to leave out until v109?

Well, I think quite a lot of things are piling up as breaking changes, so 109 will be there quite soon.

kendallb:
Perhaps for v108 the options class is marked as obsolete but remains, and we add the new property accessors for the base URL etc, then for v109 it gets removed?

kendallb:
Rebased against dev now that all the cookie code is in there. All that is left is moving authenticator to the request and the client options being hidden.

alexeyzimarev:
I think we need separate issues for moving/copying things like response writers and authenticator. I don't know exactly what those response writers are being used for.

I'd keep the authenticator property in the client options as in many (most?) cases you'd use the same authenticator for all the requests when using machine-to-machine integrations. For multi tenancy scenarios when authentication details change between the requests, it makes sense to use an authenticator scoped for a single request.

alexeyzimarev:
Aha, sorry, we already have a request-level authenticator, so the second part of my comment is irrelevant.

alexeyzimarev:
I think some things like `UserAgent` and `CachePolicy` aren't available on request level. They should not be moved, but I am fine with adding them to the request and override the client-level settings.

alexeyzimarev:
I added the CacheControl but not sure about user agent.

Issue #1882, opened 2022-07-01
Migration docs are outdated, contain false information


sirkris:
**Describe the bug**
The migration page for AddFile incorrectly states that `restRequest.AddFile((FileParameter)file);` is the valid migration for `restRequest.Files.Add((FileParameter)file);`.  However, that overload has apparently since been removed but the docs were never updated.

After spending some time on Google, I was unable to find ANYTHING mentioning the correct migration for this.  So I ended up having to create a StackOverflow thread:

https://stackoverflow.com/questions/72736161/contrary-to-restsharp-docs-request-addfile-does-not-accept-fileparameter-as-arg

Creating an extension method to restore the removed functionality seems to have worked, but I wasted a significant amount of time trying to figure out the optimal way to resolve this.

**To Reproduce**
Just try using the code above.  It seems the overload the docs say to use has since been marked as internal and is no longer accessible in userland.

**Expected behavior**
The documentation should contain accurate information.

**Stack trace**
N/A

**Desktop (please complete the following information):**
 - OS: Windows 11
 - .NET version: .NET Standard 2.0
 - 108.0.1 (migrating from 106.6.9)

**Additional context**

**I know it's a pain but PLEASE update your migration docs!!!**

PR #1888, opened 2022-07-08
Updated access modifier on RestRequest.AddFile to match documentation…

FissionTech:
… specification.

## Description
This PR aims to fix issue #1882 in which the official RestSharp documentation claims that files may be attached to a request by calling the `RestRequest.AddFile` method, which accepts a `FileParameter` as its only input. However, the internal modifier prevents this usage and instead forces the user to call any of three `AddFile` overloads in `RestSharp.RestRequestExtensions`. Each of these overloaded extension methods creates a `FileParameter` and calls the original `internal` `AddFile` method.

## Purpose
This pull request is a:

## Checklist

### DIFF
--- a/src/RestSharp/Request/RestRequest.cs
+++ b/src/RestSharp/Request/RestRequest.cs
@@ -198,5 +198,5 @@ public class RestRequest {
         return this;
     }
 
-    internal RestRequest AddFile(FileParameter file) => this.With(x => x._files.Add(file));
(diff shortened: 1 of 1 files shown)


alex-derouchie:
Can this change please be re-considered? I've hit this issue and it seems to be relatively common - it's a pretty significant inconsistency with the current documentation which directs users to make use of the method signature that is currently internal.

PR #1895, opened 2022-07-21
Correct get/post async/json issue in docs

### LINKED ISSUE #1842: Recommended Usage in docs uses GetAsync; should it be GetJsonAsync?
I'm using v107.3. I get "does not contain a definitionfor .. and no accessible extension .." compiler errors when using the example code from the docs

Example code line from [Recommended Usage](https://restsharp.dev/v107/#recommended-usage) docs says:

```
    => _client.GetAsync<GitHubRepo[]>("users/aspnet/repos");
```

Should it say `GetJsonAsync`, or do we need to import something that introduces this `GetAsync<T>(string)` method? Could that import be added to the docs?

Also same question to "simplified API" in [making requests](https://restsharp.dev/v107/#making-requests) says:

```
var request = new MyRequest { Data = "foo" };
var response = await client.PostAsync<MyRequest, MyResponse>(request, cancellationToken);
// response will be of type TResponse
```

Should it be `PostJsonAsync`/have a string URL in there?

Dwizzle252:


Dwizzle252:
Closes #1842

docs/v107/README.md
```diff
@@ -94,7 +94,7 @@ When you got a request instance, you can make a call:
 var request = new RestRequest()
     .AddQueryParameter("foo", "bar")
     .AddJsonBody(someObject);
-var response = await client.PostAsync<MyResponse>(request, cancellationToken);
+var response = await client.PostJsonAsync<MyResponse>("path/to/endpoint",request, cancellationToken);
```
alexeyzimarev, line 97: Sorry but this is wrong. The second argument for `PostJsonAsync` is not a request, it's the payload.

PR #1897, opened 2022-07-24
Add support for using models as parameter containers and add them to the parameters collection of a REST request

abenaru:
## Description
This PR introduces a feature that allows us to wrap REST request parameters under any arbitrary model and add them to a `RestRequest` instance via an extension method.

## Purpose
This pull request is a:

## Checklist

### DIFF
--- a/RestSharp.sln.DotSettings
+++ b/RestSharp.sln.DotSettings
@@ -82,6 +82,7 @@
 	<s:Boolean x:Key="/Default/UserDictionary/Words/=instagram/@EntryIndexedValue">True</s:Boolean>
 	<s:Boolean x:Key="/Default/UserDictionary/Words/=Migrator/@EntryIndexedValue">True</s:Boolean>
 	<s:Boolean x:Key="/Default/UserDictionary/Words/=Mongo/@EntryIndexedValue">True</s:Boolean>
+	<s:Boolean x:Key="/Default/UserDictionary/Words/=Parameterizer/@EntryIndexedValue">True</s:Boolean>
 	<s:Boolean x:Key="/Default/UserDictionary/Words/=repcodes/@EntryIndexedValue">True</s:Boolean>
 	<s:Boolean x:Key="/Default/UserDictionary/Words/=Serilog/@EntryIndexedValue">True</s:Boolean>
 	<s:Boolean x:Key="/Default/UserDictionary/Words/=upsert/@EntryIndexedValue">True</s:Boolean>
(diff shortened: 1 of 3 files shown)


alexeyzimarev:
Doesn't it overlap with `AddObject`?

alexeyzimarev:
I like this implementation better, and I think it could be another version of `AddObject`, which does the same by adding properties as `GetOrPost` parameters, but doesn't allow to specify the type.

abeferreira:
> I like this implementation better, and I think it could be another version of `AddObject`, which does the same by adding properties as `GetOrPost` parameters, but doesn't allow to specify the type.

Sounds good to me. Sorry for the delayed response, my schedule has been rather tight lately. I'll be working on it this weekend.

Also worth asking: What do you think of replacing the current implementation with this one as well? We could benefit from caching the instance properties for a given type and filtering them accordingly later on. I'll make sure to provide the benchmarks if that sounds good to you. This won't break the existing API, as the method signature will be left unchanged.

dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/restsharp/RestSharp?pullRequest=1897) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/restsharp/RestSharp?pullRequest=1897) before we can accept your contribution.<br/><br/>:x: oabnerferreira [sign now](https://cla.dotnetfoundation.org/restsharp/RestSharp?pullRequest=1897)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/restsharp/RestSharp?pullRequest=1897) it.</sub>

alexeyzimarev:
This time the delay is on my side, apologies. 

The problem with _replacing_ `AddObject` with a generic method is that you aren't always able to know the object type. I am sure that your implementation is much faster, but I don't think the existing method should be (or can be) replaced with this one without consequences. So, I'd rather have both of them.

abenaru:
> This time the delay is on my side, apologies.
> 
> The problem with _replacing_ `AddObject` with a generic method is that you aren't always able to know the object type. I am sure that your implementation is much faster, but I don't think the existing method should be (or can be) replaced with this one without consequences. So, I'd rather have both of them.

I see, that makes sense. Created a new PR with some changes based on this conversation. :)
https://github.com/restsharp/RestSharp/pull/1974

Issue #1899, opened 2022-07-27
Update System.Text.Json to remove a critical vulnerability


LauraArmitage:
Apologies if I've done this incorrectly, I've not reported an issue on an open source project before

Also the security pages have not been updated past the default to tell me how to report a vulnerability.

**Describe the bug**
When including RestSharp in my project the default dependencies include `System.Text.Json` at v5.0.0 which includes `System.Text.Encodings.Web` at v5.0.0

`System.Text.Encodings.Web` v5.0.0 has a security vulnerability as documented here: https://github.com/advisories/GHSA-ghhp-997w-qr28

v5.0.1 is a patched version according to that page so hopefully it's an easy, non-breaking change to update

**To Reproduce**
Create a new project and add the latest (108.0.1 as of reporting) version of RestSharp

PR #1923, opened 2022-09-13
Set minimum System.Text.Json version to 5.0.1

LauraArmitage:
## Description

Fix https://github.com/restsharp/RestSharp/issues/1899

### DIFF
--- a/src/RestSharp/RestSharp.csproj
+++ b/src/RestSharp/RestSharp.csproj
@@ -1,6 +1,6 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
     <ItemGroup Condition="$(TargetFramework) != 'net6.0'">
-        <PackageReference Include="System.Text.Json" Version="[5.0,)" />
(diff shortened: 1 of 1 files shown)


alexeyzimarev:
Thanks for the fix. One question: why have you removed the open upper version constraint?

LauraArmitage:
Sorry, I've not done this right - I was just trying to merge this into my fork. I've not tested this yet
From [this doc](https://docs.microsoft.com/en-us/nuget/concepts/package-versioning#version-ranges) I interpret that `[5.0,)` sets an inclusive minimum only and this would be the same but with a higher minimum version.

alexeyzimarev:
I think it could be `[5.0.1,)`

alexeyzimarev:
Could also be `(5.0,)`. Yes, I think your current implementation is correct, I didn't read the docs correctly.

LauraArmitage:
Thanks, I'll mark this as ready for review once I've tested

PR #1966, opened 2022-11-08
Move Cookie handling out of HttpClient so we do not cross pollinate requests

kendallb:
## Description

Implements the following:

- Move handling of Cookies out of HttpClient and into RestSharp, so they will not cross pollinate requests.

## Purpose
This pull request is a:

although the breakage is only against stuff that was not released other than in alpha anyway (making Options internal again).

## Checklist


kendallb:
Dang I was not updated to dev. Will rebase it all and fix it up now.

PR #1974, opened 2022-11-18
Introduce faster and richer way to add objects as query parameters

abenaru:
## Description

This PR is a result of the discussion in [PR#1897](https://github.com/restsharp/RestSharp/pull/1897). It introduces a new way of adding objects as query parameters to a RestRequest instance when the object type is known beforehand, which allows for a lot of optimizations to be made.

Apart from this, this PR also builds on top of pre-existing functionality, allowing for things like:
- CSV and Array format is now available for all `IEnumerable` types.
- Formatting is applied to all formattable elements in an `IEnumerable` property.
- Properties are cached and getters are compiled into a delegate using LINQ Expressions for better performance.
- If a property is not `IConvertible`, `IEnumerable` or `IFormattable`, use the type's defined `TypeConverter` and delegate conversion to string to it instead of using `.ToString()`. This allows users to create custom converters to determine how certain types should be stringified. If no custom converter is defined, we will fall back to `.ToString()` anyways.
- Ref structs are properly ignored since they cannot be boxed and don't play well with reflection, which is not taken into account in the `AddObject` implementation.

## Purpose
This pull request is a:

## Checklist


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/restsharp/RestSharp?pullRequest=1974) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/restsharp/RestSharp?pullRequest=1974) before we can accept your contribution.<br/><br/>:x: abenaru [sign now](https://cla.dotnetfoundation.org/restsharp/RestSharp?pullRequest=1974)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/restsharp/RestSharp?pullRequest=1974) it.</sub>

alexeyzimarev:
Thanks for the PR. Please remember to sign the CLA.

abenaru:
> Thanks for the PR. Please remember to sign the CLA.

Oops, my bad. Done. ✅

abenaru:
I'll take a look into what might be happening with the test case that failed once I get home.

abenaru:
The test case that was failing was one that I had in place to ensure that ref structs are not taken into consideration when turning an object into query parameters, since they cannot be boxed and will throw at runtime, which is what happens with the default `AddObject` extension method.

I added a comment explaining the fix, but basically, I was filtering out properties based on whether they had a IsByRefLikeAttribute defined, but this attribute is generated by the compiler, and different assemblies will have different versions of this attribute, which was making this check useless. So now we're just checking if the property return type has any attributes with the full name "System.Runtime.CompilerServices.IsByRefLikeAttribute" instead.

alexeyzimarev:
I had to merge it manually, for some reason GitHub doesn't recognise it.

Issue #1990, opened 2023-01-03
Restsharp adds UserAgent headers to HttpClient instance that is passed explicitly


p-skakun:
**Describe the bug**
When create multiple `RestClient` instances with the same HttpClient instance and default options it adds `UserAgent` header each time.
In certain scenarios it causes `System.InvalidOperationException: Collection was modified`

**To Reproduce**
```
var httpClient = new HttpClient();
var restClient1 = new RestClient(httpClient);
var restClient2 = new RestClient(httpClient);
Assert.AreEqual(2, httpClient.DefaultRequestHeaders.UserAgent.Count);
Assert.AreEqual(httpClient.DefaultRequestHeaders.UserAgent.First(), httpClient.DefaultRequestHeaders.UserAgent.Last());
```

**Expected behavior**
Reused httpClient keeps unique `UserAgent` headers

**Desktop (please complete the following information):**
 - OS: Windows 11 Home
 - .NET version: 7
 - Version: 109.0.0-preview.1

**Additional context**
I believe the issues starts here: 
https://github.com/restsharp/RestSharp/blob/dev/src/RestSharp/RestClient.cs#L131

In case of default `UserAgent` value `Product?.Name` is `RestSharp`, while `UserAgent` is `RestSharp/x.x.x` so the check always passes.

tacosontitan:
`ConfigureHttpClient` is not called in the constructor utilized by your reproduction steps. However, this is reproducible:

```csharp
// arrange
var httpClient = new HttpClient();
var clientOptions = new RestClientOptions();

// act
var restClient1 = new RestClient(httpClient, clientOptions);
var restClient2 = new RestClient(httpClient, clientOptions);

// assert
Assert.Contains(httpClient.DefaultRequestHeaders.UserAgent, agent => $"{agent.Product.Name}/{agent.Product.Version}" == clientOptions.UserAgent);
Assert.Single(httpClient.DefaultRequestHeaders.UserAgent);
```
In `RestClient`'s current state, this test *will* fail. This is caused by the fact that the current snippet only compares the `Product` property with the `options.UserAgent` which also contains a version.

https://github.com/restsharp/RestSharp/blob/f7d3984244c4808ae7c068e028753c2857f325b9/src/RestSharp/RestClient.cs#L128-L136

Note that line 131 here is the source of the issue.

alexeyzimarev:
Fixed by @tacosontitan in #2008, I released it in 109.0.1.

PR #2008, opened 2023-02-20
Fixed a bug causing duplicate user-agent headers when using a shared `HttpClient`.

tacosontitan:
## Description

This PR fixes issue #1990 by using string interpolation to concatenate the `Product` and `Version` properties in a way that matches the `UserAgent` property of `RestClientOptions`.

## Purpose
This pull request is a:

## Checklist


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/restsharp/RestSharp?pullRequest=2008) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/restsharp/RestSharp?pullRequest=2008) before we can accept your contribution.<br/><br/>:x: tacosontitan [sign now](https://cla.dotnetfoundation.org/restsharp/RestSharp?pullRequest=2008)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/restsharp/RestSharp?pullRequest=2008) it.</sub>

src/RestSharp/RestClient.cs
```diff
@@ -128,7 +128,7 @@ public partial class RestClient : IDisposable {
     void ConfigureHttpClient(HttpClient httpClient) {
         if (Options.MaxTimeout > 0) httpClient.Timeout = TimeSpan.FromMilliseconds(Options.MaxTimeout);
 
-        if (Options.UserAgent != null && httpClient.DefaultRequestHeaders.UserAgent.All(x => x.Product?.Name != Options.UserAgent)) {
+        if (Options.UserAgent != null && httpClient.DefaultRequestHeaders.UserAgent.All(x => $"{x.Product?.Name}/{x.Product?.Version}" != Options.UserAgent)) {
```
alexeyzimarev, line 131: Good catch :) How about moving the default user agent to a get-only private property, so we have this calculated in one place? The code which sets the header and checks if it's already there will use the same value even if we decide to change how it's composed.

tacosontitan: I'm not sure I fully understand your idea, and I believe it's because I'm tripping myself up mentally due to the fact that the default user agent is already a `static readonly` field:

https://github.com/restsharp/RestSharp/blob/f7d3984244c4808ae7c068e028753c2857f325b9/src/RestSharp/RestClientOptions.cs#L26-L28

Are you proposing that we encapsulate the user agent so that the composition and equality checks are handled from one place? I can certainly implement such a feature, I just want to make sure I'm comprehending your request before I get started.

Can you elaborate for clarity?

alexeyzimarev: Yeah, I haven't looked. Even better. I mean replacing
```
$"{x.Product?.Name}/{x.Product?.Version}"
```
with
```
DefaultUserAgent
```

tacosontitan: A few questions:

1. Are you comfortable with the encapsulated approach?
2. Should we handle this in `v109` as well?
3. If yes to both, I must create a separate pull request for `v109` since the `RestSharpOptions` file is stored in that branch's `Options` folder. I recommend that we create the encapsulation at the root where `RestSharpOptions` can be found in `dev` so at least that part merges smoothly, but the implementation in `RestSharpOptions` would need to be implemented by hand in both.

I'm ready to start working on this locally.

alexeyzimarev: It's fine to merge to main as it won't introduce conflicts with v109. There's no need to create a separate PR for v109 branch.

Please ignore the whole conversation here, I didn't realise what your code is doing and I get it now. I think it's great, please change nothing. It's all correct. Sorry about the confusion.

tacosontitan:
I'm not seeing why the test is exiting with code 1. I'm more than willing to make adjustments if needed. Can someone help me identify what's going wrong there?

tacosontitan:
Nevermind, I found it. Failed 1 test, it was hidden away at first. I'll review and update once resolved.

src/RestSharp/RestClient.cs
tacosontitan:
```csharp
using System.Net.Http.Headers;

namespace RestSharp;

/// <summary>
/// Represents a user agent for RestSharp.
/// </summary>
public sealed class RestSharpUserAgent
{
    private static readonly string _version = typeof(RestClientOptions).Assembly.GetName()?.Version?.ToString() ?? "0.0.0.0";
    /// <summary>
    /// Initializes a new instance of the <see cref="RestSharpUserAgent"/> class.
    /// </summary>
    /// <param name="name">The name of the user agent.</param>
    /// <param name="version">The version of the user agent.</param>
    public RestSharpUserAgent(string name, string version) {
        Name = name;
        Version = version;
    }
    /// <summary>
    /// The name of the user agent.
    /// </summary>
    public string Name { get; }
    /// <summary>
    /// The version of the user agent.
    /// </summary>
    public string Version { get; }
    /// <summary>
    /// The default <see cref="RestSharpUserAgent"/>.
    /// </summary>
    /// <returns></returns>
    public static RestSharpUserAgent Default => new RestSharpUserAgent("RestSharp", _version);
    /// <summary>
    /// Returns a <see cref="string"/> representing the name and version of the <see cref="RestSharpUserAgent"/>.
    /// </summary>
    /// <remarks>This will return <see cref="Name"/>/<see cref="Version"/>.</remarks>
    /// <returns>A <see cref="string"/> representing the name and version of the <see cref="RestSharpUserAgent"/>.</returns>
    public override string ToString() => $"{Name}/{Version}";
    /// <summary>
    /// Determines whether the specified <see cref="object"/> is equal to the current <see cref="RestSharpUserAgent"/>.
    /// </summary>
    /// <param name="obj">The <see cref="object"/> to compare the current <see cref="RestSharpUserAgent"/> with.</param>
    /// <returns><see langword="true"/> if the <see cref="object"/> is equal to the current <see cref="RestSharpUserAgent"/>, otherwise <see langword="false"/>.</returns>
    public override bool Equals(object obj)
    {
        if (obj is string objString)
            return objString.Equals(ToString());

        if (obj is RestSharpUserAgent agent)
            return agent.Name.Equals(Name) && agent.Version.Equals(Version);

        if (obj is ProductInfoHeaderValue productHeader)
            return productHeader.Product.Name.Equals(Name) && productHeader.Product.Version.Equals(Version);

        return obj.Equals(this);
    }
    public static implicit operator string(RestSharpUserAgent agent) => agent.ToString();
    public static implicit operator RestSharpUserAgent(string agent)
    {
        if (agent is null || string.IsNullOrWhiteSpace(agent) || !agent.Contains('/'))
            return Default;

        IEnumerable<string> split = agent.Split('/');
        if (split.Count() != 2)
            return Default;

        return new RestSharpUserAgent(split.First(), split.Last());
    }
}
```

There's no discussion area, so I can't post what I had already there. Here's the start of an encapsulated approach, in-case the project ever goes in that direction.

tacosontitan:
After local review, the test in question fails in the `dev` branch without my change. I tried reviewing the action history for confirmation, but that yielded no results so I must leave it to the project maintainers at this point.

src/RestSharp/RestClient.cs
alexeyzimarev: You can just use a record to avoid explicit equality implementation.

alexeyzimarev:
Deprecated by #2010

tacosontitan:
@alexeyzimarev, I reviewed the files changed in #2010 and don't see where this was accounted for. Was the correct item linked?

alexeyzimarev:
Ah sorry. I might have messed up. Let me check it now. Apologies!

alexeyzimarev:
@tacosontitan Cloud you check the message from the CLA bot? It's a new thing from .NET Foundation, would be nice if you can do what it asks.

tacosontitan:
@dotnet-policy-service agree

tacosontitan:
@alexeyzimarev, no problem, also had to update the PR with the latest changes in `dev`.

PR #2039, opened 2023-03-29
Make BuildUrl an extension

### LINKED ISSUE #2022: OAuth1 AddParameter/AddQueryParameter Empty Value Gets "Value Cannot Be null" Exception
**DO NOT USE ISSUES FOR QUESTIONS**

Describe the bug
Api URL - https://example.com/aps/2/collections/accounts?eq(id,1002305)
Authentication - OAuth1

when trying to POST\GET requests with empty value when using **OAuth1** Authentication 
getting "Value Cannot Be null"  Exception

To Reproduce
POST\GET Request to https://example.com/aps/2/collections/accounts?eq(id,1002305)
with
Request.AddParameter("eq(id,1002305)",null, false);
or
Request.AddQueryParameter("eq(id,1002305)",null, false);

Expected behavior
manage to POST\GET with AddParameter\AddQueryParameter with null value using **OAuth1** Authentication

Desktop (please complete the following information):

OS: Windows10
.NET 7
Version 109.0.1

image

image

### LINKED ISSUE #2025: DefaultParameters collection is not read-only
The `DefaultParameters` property of `RestClient` has the `ParametersCollection` type. Although it's a read-only collection, it exposes public methods to add and remove parameters.

When using `RestClient.AddDefaultParameter` method, the collection is updated in a locked scope. However, it is possible to manipulate the collection directly by calling `client.DefaultParameters.AddParameter`, and it will be execute without locking the collection. It basically makes the client not thread-safe if default parameters are updated this way. It's not an intended use of the collection, but it has a public API, which allows people to do that.

My suggestion is to make the `DefaultParameters` property internal or private. The `AddDefaultParameter` might need to be added to the interface.

### LINKED ISSUE #2026: Make `BuildUri` public
From the previous discussion:

The only property not including in the interface is `DefaultParameters`. 
If this will not be a part of the interface or of the IRestClientOptions it could be optional:
```
public static Uri BuildUri(this IRestClient restClient, RestRequest request, ParametersCollection? defaultParameters = null) {
    defaultParameters   ??= new ParametersCollection();
    var (uri, resource) =   restClient.Options.BaseUrl.GetUrlSegmentParamsValues(request.Resource, restClient.Options.Encode, request.Parameters, defaultParameters);
    var mergedUri = uri.MergeBaseUrlAndResource(resource);

    var finalUri = mergedUri.ApplyQueryStringParamsValuesToUri(
        request.Method,
        restClient.Options.Encoding,
        restClient.Options.EncodeQuery,
        request.Parameters,
        defaultParameters
    );
    return finalUri;
}
```
If this method will not be accessible in future it is bad for me but solvable.

_Originally posted by @nesc58 in https://github.com/restsharp/RestSharp/issues/1976#issuecomment-1337533130_

### LINKED ISSUE #2037: No charset in url-encoded form in RestSharp 109.0.1
**Describe the bug**
no cookie in the restresponse cookies containter with .NET 7.0 and RestSharp 109.0.1

**To Reproduce**
```
var options109 = new RestClientOptions($"{realm.ConfigSource}")
{
         RemoteCertificateValidationCallback = (restclientsender, certificate, chain, sslPolicyErrors) => true,
         MaxTimeout = -1,
};

var restclient = new RestClient(options109);

var loginrequest = new RestRequest();
                
loginrequest.Resource = "login";
loginrequest.Method = Method.Post;
loginrequest.AddParameter("username", realm.Username);
loginrequest.AddParameter("password", realm.Password);
               
var loginresult = await restclient.ExecuteAsync(loginrequest).ConfigureAwait(false);
               
```

The exact same code returns a cookie on .NET 6 and RestSharp 108.0.3

**Stack trace**
no exception or any errors

**Desktop (please complete the following information):**
 - OS: Azure Function
 - .NET version .NET 7.0
 - Version 109.0.1

**Additional context**
The differences in the restresponse between the two versions:

108 
ContentHeaders: (there is only 1, and the ContentType is empty)
{HeaderParameter { Name = Content-Type, Value = text/html; charset=utf-8, Type = HttpHeader, Encode = False, ContentType =  }}

ContentLength = null
RawBytes = {byte[695223]}

109 
ContentHeaders: (there are 1, and the ContentType is text/plain)
{HeaderParameter { Name = Content-Type, Value = text/html; charset=utf-8, Type = HttpHeader, Encode = False, ContentType = text/plain }}
{HeaderParameter { Name = Content-Length, Value = 2352, Type = HttpHeader, Encode = False, ContentType = text/plain }}

ContentLength = 2352
RawBytes = {byte[2352]}

I need to use .NET 7.0 and on that platform:
108.0.3 has a bug and throws an exception explained here: [https://github.com/restsharp/RestSharp/issues/1969](url)
109.0.1 doesn't work with the above issue. 

If someone gives me a solution/workaround I'll happily contribute $50 for this project, because currently I can't do my work as nothing works since the .NET 7.0 upgrade.

alexeyzimarev:
- Split the type for request parameters and default parameters
- Default parameters collection locks for concurrent updates (fixed #2025)
- Add default parameters to the interface
- Make `BuildUrl` a public extension (fixes #2026)
- Change authenticator interface to use `IRestClient`

- Also fix the missing charset for url-encoded forms (fixes #2037)

- Allow null parameter value in OAuth1 authenticator (fixes #2022)


nesc58:
RestSharp should be versioned with semantic versioning.

Changing ParametersCollection is a breaking change
```new ParametersCollection(...)``` is not longer available. It must be replaced with `new DefaultParameters(...)` or `new RequestParameters(...)`.

The changes are fine to me and reasonable but the version should be changed to `110.0.0`. This will make the breaking change clear because it is not only a bug fix or added functionality. e.g. I am using version patterns like
```
<PackageVersion Include="RestSharp" Version="[109.0.1, 110.0.0)" />
```
to prevent other libraries installing a potential breaking version. The current change will break this pattern.

Is this fine for you to change the version to `110.0.0`?

alexeyzimarev:
Yes, the plan is to release the new major.

PR #2042, opened 2023-04-01
Adding client-level cookie container

alexeyzimarev:
* Added a custom cookie container to `RestClientOptions`
* It _will not_ be added to the message handler
* Cookies will be added as headers the same way as request cookies

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -150,6 +150,30 @@ services.AddSingleton<ITwitterClient>(
 );
 ```
 
+### Simple factory
+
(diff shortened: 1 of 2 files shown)


kendallb:
My concern would be adding the cookies to the container in the request because if the user is using a request specific container that they are maintaining, it means the client ones will end up in that container after the request. 

cookieContainer.Add(Options.CookieContainer.GetCookies(url));

Is it feasible to simply call headers.AddCookieHeaders(cookieContainer, url) for both cookie containers rather than combined them?

alexeyzimarev:
Yeah, I thought about that as one of the options, but I haven't considered the request container concern. You are right, will change it now.

alexeyzimarev:
@kendallb I updated it following your comment.

kendallb:
Looks good to me

PR #2043, opened 2023-04-03
Add `AddJsonBody` overload to serialise top-level string

### LINKED ISSUE #1954: AddJsonBody handles strings in a usually user-friendly but incorrect way
**Describe the bug**

When calling `AddJsonBody()` with a string, what is added is the literal string supplied without serialization. This makes it impossible to handle the case where the value is a JSON object whose full and complete structure is `"String"`.

**To Reproduce**

The RestSharp code involved is trivial, see the body of `AddJsonBody()`:

```csharp
        request.RequestFormat = DataFormat.Json;
        return obj is string str ? request.AddStringBody(str, DataFormat.Json) : request.AddParameter(new JsonParameter("", obj, contentType));
```

In our case, the API we're hitting has this type stanza:

```json
"requestBody": {
          "content": {
            "application/json": {
              "schema": {
                "type": "string"
              }
            }
          }
        },
```

Importantly, the backend expects this input to be the serialized version of a JSON object (I know, I know). So to call it properly, naively we expected to be able to do:

```csharp
  request.AddJsonBody(JsonConvert.SerializeObject(myObject))
```

But of course this does not work -- it hits the first part of the ternary above and treats it as "already serialized", sending the string as-is. This causes it to be parsed as an object by the middleware, not as a string as intended.

Instead, this is currently what we have to do:

```csharp
  request.AddJsonBody(JsonConvert.SerializeObject(JsonConvert.SerializeObject(myObject)))
```

(Note that in the real world it's not as simple as this -- we are calling RestSharp through a generated client, so we don't have the option of calling a different more appropriate method even if there is one.)

**Expected behavior**
There should be a way to add a top-level string as JSON, serializing it as expected. Top-level strings are valid JSON!

I understand that the current implementation probably avoids some common gotchas, but it is important to handle this case as well, because there really are APIs in the wild that require JSON-formatted top-level strings as input. In our case, this issue required a lot of debugging to figure out.

alexeyzimarev:
## Description

Proposal to fix #1954.

Add a `AddJsonBody` overload with `string` payload argument with `forceSerialize` argument. If `forceSerialize` is set to true, the string will be serialised. If it's set to `false`, the string will be sent as-is. 

## Purpose

It's not a strictly breaking change, I think it might be released a s a minor version.

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -339,15 +339,35 @@ When you call `AddJsonBody`, it does the following for you:
 - Sets the content type to `application/json`
 - Sets the internal data type of the request body to `DataType.Json`
 
-::: warning
-Do not send JSON string or some sort of `JObject` instance to `AddJsonBody`; it won't work! Use `AddStringBody` instead.
-:::
-
 Here is the example:
 
 ```csharp
 var param = new MyClass { IntData = 1, StringData = "test123" };
 request.AddJsonBody(param);
+```
+
+It is possible to override the default content type by supplying the `contentType` argument. For example:
+
+```csharp
+request.AddJsonBody(param, "text/x-json");
+```
+
+If you use a pre-serialized string with `AddJsonBody`, it will be sent as-is. The `AddJsonBody` will detect if the parameter is a string and will add it as a string body with JSON content type.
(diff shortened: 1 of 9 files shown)


d-a-s:
I'd recommend you consider adding this as a breaking change in the [release notes](https://github.com/restsharp/RestSharp/releases/tag/110.1.0).

I recently upgraded an old project that was relying on the previous top level string behavior (encoding by default).

PR #2048, opened 2023-04-04
feature: Added Option to add Interceptors

fseidl-bauradar:
## Description

## Purpose
This pull request is a:

## Checklist


alexeyzimarev:
Looks good. One comment: why interfaces? These are simple functions, would it not be sufficient to use named delegate types instead? Interfaces are high-cost abstraction compared with functions.

fseidl-bauradar:
@dotnet-policy-service agree company="Bauradar"

fseidl-bauradar:
> Looks good. One comment: why interfaces? These are simple functions, would it not be sufficient to use named delegate types instead? Interfaces are high-cost abstraction compared with functions.

I just like Interfaces more and additionally thought about future usage, where a Request/Response Pair is handled in both directions

PR #2049, opened 2023-04-04
 Added Option to add Interceptors

fseidl-bauradar:
## Description

## Purpose
This pull request is a:

## Checklist


fseidl-bauradar:
> The test project is in the wrong place and doesn't have the project file. I don't think it has any tests either.

I didn't changed the possition, as far as I see there are sub Projects inside the test folder.
Which contain the tests.

alexeyzimarev:
Also, those interfaces can't be used alone as you can't add, for example, a class that implements `IBeforeSerializationInterceptor` to the interceptors list because the class must implement the `IInterceptor` interface. Based on that, I think it would be easier to have a base abstract class (that's how gRPC interceptors work) with empty virtual methods that can be overridden. With the current implementation you'd need to implement all the methods even if you want to intercept only one thing.

alexeyzimarev:
> > The test project is in the wrong place and doesn't have the project file. I don't think it has any tests either.
> 
> I didn't changed the possition, as far as I see there are sub Projects inside the test folder. Which contain the tests.

Here's how the `Tests` solution folder looks like:

image

And the `ConsoleTest` directory is on the top level, it has no tests and no project file:

image

fseidl-bauradar:
> > > The test project is in the wrong place and doesn't have the project file. I don't think it has any tests either.
> > 
> > 
> > I didn't changed the possition, as far as I see there are sub Projects inside the test folder. Which contain the tests.
> 
> Here's how the `Tests` solution folder looks like:
> 
> image
> 
> And the `ConsoleTest` directory is on the top level, it has no tests and no project file:
> 
> image

Oh seems like there is a little to much in the Pull Request, I originally wanted to exclude the ConsoleTest Project, from Repository.
It was just a little playgroud for setting up and test if the Interceptor is working as expected

fseidl-bauradar:
> I don't think you needed to change this file:
> 
> image

You are right, but Rider mentioned it, and I aggree with it. As far I know Classes should always start with UpperCase

PR #2052, opened 2023-04-04

fseidl-bauradar:
## Description

## Purpose
This pull request is a:

## Checklist


src/RestSharp/Interceptors/IBeforeSerializationInterceptor.cs
```diff
@@ -0,0 +1,14 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace RestSharp.Interceptors {
+    public interface IBeforeSerializationInterceptor {
```
alexeyzimarev, line 6: I still don't understand why these interfaces are needed as they can't be used. I am also not sure what's the purpose of the `IInterceptor` interface.

fseidl-bauradar: The original purpose of this Interfaces was to make the Interceptor compatible with your delegates, so the three Interceptors can be assigned to the Request to.

In the first step, I thought about combining the handling of the Request Delegates with the handling of the Interceptors.

var List<IBeforeRequestInterceptor> beforeRequestInterceptors = new();
beforeRequestInterceptors.add(request.OnBeforeRequest);
beforeRequestInterceptors.add(Options.Interceptors);
foreach(interceptor in beforeInterceptors){
    await interceptor.InterceptBeforeRequest(request);
}

fseidl-bauradar: class BeforeRequestInterceptor: IBeforeRequestInterceptor {
...
}
var beforeRequestInterceptor = new BeforeRequestInterceptor();
request.OnBeforeRequestInterceptor += beforeRequestInterceptor;

alexeyzimarev: > Another reason why I added the IInterceptor Interface, was that I don't like adding Classes as Parameters, I think it's always better to use Interfaces instead.

Why? I don't think code is about what we like or don't. Each line of code potentially adds complexity and certainly adds maintenance. Using interfaces is one of the ways in .NET for creating abstractions, and if you don't plan for the abstraction to be used, there's no need for it.

Regarding separate collections for different interjectors, I'd prefer making the existing ones obsolete and just remove them later. I don't want to keep redundant functionality, and interceptors are indeed much cleaner and more idiomatic way to do the same thing as those delegates are doing now.

fseidl-bauradar: For sure, each Line of Code adds Complexity, but, Interfaces are just a promise what a implementing class can do,
There is no working Code inside them.

Interfaces provide the Option to create a complete own implementation

If you create a Interface you know that there is nothing done in the implementation, but if you derive from a super class, there can be Implementations that disturb your way of working with them.

Add multiple Interfaces to a class but only one super class.

fseidl-bauradar: Is there anyway that we can get this Changes in the offizial Repository?

fseidl-bauradar:
Because there have been commits in the Branch and I need to synchronize

PR #2058, opened 2023-04-07
Follow redirects

### LINKED ISSUE #2059: [RFC] Implement follow redirects
The issue of redirect follow not working because of the missing cookies (#2045) triggered the following thought.

The default redirect follow by the message handler has these limitations:
- Cookie headers get populated to the consequent request only if the cookie container is used. When we removed the cookie container from the client in v109, this scenario stopped working (described in the issue referenced above)
- Redirect won't be followed when changing the scheme
- Max redirect count isn't supported in WASM
- Need to test if `SeeOther` redirect is supported, it might not be
- Need to test if headers are passed to the redirect

I did an experiment here https://github.com/restsharp/RestSharp/pull/2058 to check feasibility, and it seems to be working fine. Everything described above can be implemented, some of the things need to be guarded with opt-in options (secure to insecure redirect support, forward auth header, maybe something else).

Any thoughts and suggestions welcome.

alexeyzimarev:
## Description

Fixes #2059

Implement custom follow redirect functionality instead of relying on the message handler as it has limitations described in the issue.

## Purpose

This pull request is a:

### DIFF
--- a/src/RestSharp/Request/RestRequestExtensions.cs
+++ b/src/RestSharp/Request/RestRequestExtensions.cs
@@ -338,7 +338,7 @@ public static class RestRequestExtensions {
                 DataFormat.Json   => request.AddJsonBody(obj, contentType),
                 DataFormat.Xml    => request.AddXmlBody(obj, contentType),
                 DataFormat.Binary => request.AddParameter(new BodyParameter("", obj, ContentType.Binary)),
-                _                 => request.AddParameter(new BodyParameter("", obj.ToString(), ContentType.Plain))
+                _                 => request.AddParameter(new BodyParameter("", obj.ToString()!, ContentType.Plain))
             };
         }
 
--- a/src/RestSharp/RestClient.Async.cs
+++ b/src/RestSharp/RestClient.Async.cs
@@ -86,16 +86,11 @@ public partial class RestClient {
             throw new ObjectDisposedException(nameof(RestClient));
         }
 
-        using var requestContent = new RequestContent(this, request);
-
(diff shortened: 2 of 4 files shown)


rassilon:
I've added a POST -> GET redirect with recieving cookies during the original 302 redirect response test to #2119.

rassilon:
@alexeyzimarev , would you like me to try doing a set of your and my changes for follow redirects on top of the merged PR #2056?

alexeyzimarev:
@rassilon I actually didn't check the default implementation in the socket message handler, thanks for the links. My idea was more or less to follow, at least from settings perspective, the Flurl path. They have things like allow redirecting from https to http, something that just doesn't work with HttpClient. I haven't looked at your PR yet, will check soon, and big thanks for this contribution. It would be great to resolve the redirect issue.

rassilon:
Ahh.. I see what you mean by Flurl.. very neat idea. Would you want to add settings for each of the decisions in the socket message handler? Or support the simpler approach Flurl uses? Which is apparently at: https://github.com/tmenier/Flurl/blob/dev/src/Flurl.Http/FlurlClient.cs#L162 atm. (i.e. a simple settings object and an event handler)

rassilon:
I can give a whirl at making some more progress on this next week if you have other higher priority fish to fry.

alexeyzimarev:
> simpler approach Flurl uses

We can't use the socket message handler anyway as it's only .NET 5+, or wrap the handler instantiation in pragma (also an option). But anyway, yes, I just want to make own decisions as I mentioned https://github.com/restsharp/RestSharp/issues/2059#issuecomment-1685235640

Issue #1422, opened 2020-01-31
Streaming files to API (uploading) without multipart/form-data


maeni70:
## Expected Behavior
Allow sending files in chunks (via stream) without multipart/form-data.

The same question has already been asked at stackoverflow:
https://stackoverflow.com/questions/35478663/upload-file-without-multipart-form-data-using-restsharp

However, it has not been correctly answered.

Is there any way how to stream large files to an API but without multipart/form-data? I retrieve an "UnsupportedMediaType" response when trying to do so. It seems that RestSharp is automatically sending the request as multipart/form-data if files are attached in body.

There are examples how to do with .NET HttpWebRequest (see linked stackoverflow question above) or Java, but I could not achieve with RestSharp library.

Thanks,
Roger

maeni70:
For the specific upload functionality I have written own logic using standard HttpClient, which is a pity as I have to maintain an additional client object (with corresponding redundancies).

PR #2068, opened 2023-04-22
Support file uploading without multipart/form-data

RomanSoloweow:
Support file uploading without multipart/form-data based on #1422

### DIFF
--- a/src/RestSharp/Request/RequestContent.cs
+++ b/src/RestSharp/Request/RequestContent.cs
@@ -38,36 +38,30 @@ class RequestContent : IDisposable {
 
     public HttpContent BuildContent() 
     {
-     if (_request.AlwaysMultipartFormData && _request.AlwaysSingleFileAsContent) 
-        throw new ArgumentException("Failed to put file as content because flag AlwaysMultipartFormData enabled");
-                
-        var postParameters       = _parameters.GetContentParameters(_request.Method).ToArray();
+        var postParameters = _parameters.GetContentParameters(_request.Method).ToArray();
         var postParametersExists = postParameters.Length > 0;
         var bodyParametersExists = _request.TryGetBodyParameter(out var bodyParameter);
-        var filesExists          = _request.Files.Any();
-        
+        var filesExists = _request.Files.Any();
+
         if (filesExists)
-            AddFiles(postParametersExists, bodyParametersExists);
-        
+            AddFiles();
+
         if (bodyParametersExists)
             AddBody(postParametersExists, bodyParameter!);
-        
+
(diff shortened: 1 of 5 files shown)


src/RestSharp/Request/RequestContent.cs
```diff
@@ -36,38 +36,66 @@ class RequestContent : IDisposable {
...
     }
 
-    public HttpContent BuildContent() {
-        AddFiles();
-        var postParameters = _parameters.GetContentParameters(_request.Method).ToArray();
-        AddBody(postParameters.Length > 0);
-        AddPostParameters(postParameters);
+    public HttpContent BuildContent() 
+    {
+     if (_request.AlwaysMultipartFormData && _request.AlwaysSingleFileAsContent)
```
alexeyzimarev, line 41: Maybe it would be good to fail fast by moving the mutual exclusivity check to the property setters?

src/RestSharp/Request/RequestContent.cs
```diff
@@ -36,38 +36,66 @@ class RequestContent : IDisposable {
...
-        }
-
-        Content = mpContent;
+    
+     void AddFiles(bool postParametersExists, bool bodyParametersExists) 
+     {
+         // File uploading without multipart/form-data
+         if (_request.AlwaysSingleFileAsContent && _request.Files.Count == 1)
+         {
+             if (postParametersExists) throw new ArgumentException("Failed to put file as content because added post parameters");
```
alexeyzimarev, line 68: Just got an idea that we can create some sort of request validator in the request class itself, and it can be called each time anything gets changed there. Then, these rules will be in the request instead of checking it in many places.

alexeyzimarev:
Also we need a test for it.

src/RestSharp/Request/RequestContent.cs
RomanSoloweow: Calling validation in different setters will complicate the readability of the code. And the validation of adding parameters is difficult to implement because the user can use ``` request.Parameters.AddParameter ```

src/RestSharp/Request/RequestContent.cs
RomanSoloweow: Agreed, but doing validation in the Execute method seems like a good solution

RomanSoloweow:
@alexeyzimarev can you check please?

RomanSoloweow:
@alexeyzimarev ping

alexeyzimarev (approved):

alexeyzimarev:
Sorry for the delay, some unrelated tests are failing. I will see if I can fix those.

PR #2069, opened 2023-04-23
No value OAuth1 still need "=" to be valid

ili101:
## Purpose
This pull request is a:

## Checklist

In 110.0.0 a fix for https://github.com/restsharp/RestSharp/issues/2022 was added.
While the request now being sent successfully the signature generated is invalid. This is as in OAuth1 [the name is separated from the corresponding value by an '=' character (ASCII code 61), even if the value is empty](https://oauth.net/core/1.0a/#anchor13)  
So this removes the new logic to add the "=" only if value in not null (only from the OAuth1 signature not for the URL construction)


ili101:
@dotnet-policy-service agree

alexeyzimarev (approved):

ili101:
Reopen as was closed by bot?
BTW we are using this branch for some time by now and it works nicely for us

PR #2076, opened 2023-05-04

fseidl-bauradar:
## Description

## Purpose
This pull request is a:

## Checklist


fseidl-bauradar:
https://github.com/restsharp/RestSharp/pull/2052

According to the last conversation, you don't like the interfaces, so now I removed them and moved all up into a abstract class named Interceptor, which has 4 Methodes which can be overriden.

fseidl-bauradar:
No activity by the author at all

fseidl-bauradar:
I don't see how my code could produce this error.

fseidl-bauradar:
But I found out that exceptions thrown inside the Interceptor are not propagated as expected.

OnBeforeRequest was inside a non async Context, so I moved the call out including code before.
OnBeforeDeserialize was inside a non async context and inside a try catch block, because of this I moved them out
OnAfterRequest is inside a catch block,  I moved it out.

fseidl-bauradar:
Closed because of correction

fseidl-bauradar:
Locally all tests but 3 are running successful

grafik

No interaction with Interceptors or any change that would explain this, I thing that this lies in the realm of the EchoHandler.
Additionally, it's important that the test with the unknown, host checks Language specific Exception Text, this can be a problem for people running test on non english machines

PR #2078, opened 2023-05-06
Timeout as TimeSpan, Support custom request timeout

RomanSoloweow:
1. I think the current approach to use int for timeout is outdated and it generates a lot of errors when you need to remember that these are milliseconds, and you need translate it. Even HttpClient has timeout as TimeSpan.

2. The presence of "MaxTimeout" also confuses many users. If you want to set a custom timeout for the request, you need to remember which timeout is set for the rest client. Often, when you set a custom timeout for request, it's higher timeout than the default. These are requests for downloading and uploading files. It's rare case when the timeout is specially reduced for a particular request.

### DIFF
--- a/src/RestSharp/Options/RestClientOptions.cs
+++ b/src/RestSharp/Options/RestClientOptions.cs
@@ -172,10 +172,9 @@ public class RestClientOptions {
     public CookieContainer? CookieContainer { get; set; }
 
     /// <summary>
-    /// Maximum request duration in milliseconds. When the request timeout is specified using <seealso cref="RestRequest.Timeout"/>,
-    /// the lowest value between the client timeout and request timeout will be used.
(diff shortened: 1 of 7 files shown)


RomanSoloweow:
@alexeyzimarev  

tests failed again. Pull request does not affect this.
image

RomanSoloweow:
@alexeyzimarev ping

RomanSoloweow:
@alexeyzimarev ping

alexeyzimarev:
I am not merging it because it will break things. Still thinking how to make it available with some transition.

RomanSoloweow:
@alexeyzimarev  any news?

alexeyzimarev:
I noticed that one and flagged it, but for some reason I didn't think that `ConfigureHttpClient` is called when an external instance is used. 

I will release a fix (the code is already fixed).

I think `ConfigureHttpClient` needs to be removed as all what's left there is setting `Expect100Continue` and it's a header.

PR #2108, opened 2023-06-28

fseidl-bauradar:
## Description

## Purpose
This pull request is a:

## Checklist


RestSharp.sln
```diff
@@ -35,7 +35,7 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "RestSharp.Tests.Serializers
 EndProject
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "SourceGen", "SourceGen", "{55B8F371-B2BA-4DEE-AB98-5BAB8A21B1C2}"
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "SourceGenerator", "gen\SourceGenerator\SourceGenerator.csproj", "{FE778406-ADCF-45A1-B775-A054B55BFC50}"
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "SourceGenerator", "gen\SourceGenerator\SourceGenerator.csproj", "{FE778406-ADCF-45A1-B775-A054B55BFC50}"
```
alexeyzimarev, line 38: Changes to this file aren't required. Also, adding Appveyor-specific build targets need to be removed.

src/RestSharp/RestClient.Async.cs
```diff
@@ -95,38 +96,67 @@ public partial class RestClient {
...
-                .AddCookieHeaders(url, cookieContainer)
-                .AddCookieHeaders(url, Options.CookieContainer);
-
-            message.AddHeaders(headers);
-
-            if (request.OnBeforeRequest != null) await request.OnBeforeRequest(message).ConfigureAwait(false);
-
-            var responseMessage = await HttpClient.SendAsync(message, request.CompletionOption, ct).ConfigureAwait(false);
-
+            responseMessage = await HttpClient.SendAsync(message, request.CompletionOption, ct).ConfigureAwait(false);
```
alexeyzimarev, line 116: What was the reason to move the declaration of `responseMessage` to line 100?

src/RestSharp/RestSharp.csproj
```diff
@@ -39,4 +39,7 @@
     <ItemGroup>
         <ProjectReference Include="$(RepoRoot)\gen\SourceGenerator\SourceGenerator.csproj" OutputItemType="Analyzer" ReferenceOutputAssembly="false" />
     </ItemGroup>
+    <ItemGroup>
```
alexeyzimarev, line 42: It doesn't sound right, adding this package.

test/RestSharp.InteractiveTests/RestSharp.InteractiveTests.csproj
```diff
@@ -8,4 +8,7 @@
         <ProjectReference Include="$(RepoRoot)\src\RestSharp\RestSharp.csproj" />
         <ProjectReference Include="..\RestSharp.Tests.Shared\RestSharp.Tests.Shared.csproj" />
     </ItemGroup>
+    <ItemGroup>
+      <PackageReference Include="xunit.assert" Version="2.4.2" />
```
alexeyzimarev, line 12: This project doesn't change, why did you add a reference?

test/RestSharp.Tests.Serializers.Xml/SampleClasses/twitter.cs
```diff
@@ -105,7 +105,7 @@ public class complexStatus {
     public string in_reply_to_screen_name { get; set; }
 
     // ignore contributors for now
-    [DeserializeAs(Name = "user.following")]
+    [DeserializeAs(Name = "User.following")]
```
alexeyzimarev, line 108: That's wrong as the message format is camel case.

alexeyzimarev:
@fseidl-bauradar I am ready to merge this if you find time to address my comments.

src/RestSharp/Options/RestClientOptions.cs
```diff
@@ -64,6 +65,8 @@ public class RestClientOptions {
     /// </summary>
     public IAuthenticator? Authenticator { get; set; }
 
+    public List<Interceptor> Interceptors { get; set; } = new();
```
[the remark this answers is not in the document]
fseidl-bauradar: Ok, that's a point I didn't had in mind, is there a way to allow adding a interceptor after creating a inititing the client.

Currently I am using a Interceptor to refresh the jwt Token if it's outdated. For this I need a RestClient to access the backend over a Service.

src/RestSharp/RestClient.Async.cs
fseidl-bauradar: The reason, was that I handle the response inside the interceptor and prevent the tryCatch-Block from, catching exceptions thrown inside the interceptor.
For example I use an interceptor, to preprocess ErrorMessage to convert them back into Exceptions.

test/RestSharp.Tests.Serializers.Xml/SampleClasses/twitter.cs
fseidl-bauradar: This, was just a thing I found, I just changed the name of the User-Class to starting UpperCase, according to the C# Convention for Classes

alexeyzimarev: Yeah, I get it. Have the same urge to fix things sometimes. But, it is better to keep the PR scope to its purpose as things might get break unexpectedly.

test/RestSharp.InteractiveTests/RestSharp.InteractiveTests.csproj
fseidl-bauradar: Think this happed during checking, why the failing tests not working, wanted to find the reason for this.

src/RestSharp/Options/RestClientOptions.cs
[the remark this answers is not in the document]
fseidl-bauradar: But doesn't mean this that it is impossible to add an interceptor after initiating the client

RestSharp.sln
fseidl-bauradar: Seem like my ide is doing this by itself, I will try to keep this out from git in the future

test/RestSharp.Tests.Serializers.Xml/SampleClasses/twitter.cs
fseidl-bauradar: We can let it unchanged if you want, it's not as important for me

.gitignore
```diff
@@ -53,4 +55,5 @@ RestSharp.IntegrationTests/config.json
 /out/
 /docs/.vuepress/dist/
 .vscode/
-.temp/
\ No newline at end of file
+.temp/
+global.json
```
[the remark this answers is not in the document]
fseidl-bauradar: Ok, I will keep this in mind. I thought that it would be a IDE config file

src/RestSharp/RestSharp.csproj
fseidl-bauradar: I think, we can remove it, seems like I added it by accident, is still building after removing it.

src/RestSharp/Options/RestClientOptions.cs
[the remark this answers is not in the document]
fseidl-bauradar: Ok, I checked the Microsoft Reference, as I understand now you can not change the list directly, but you can change the list that is wrapped by ReadOnlyCollection.
But how do we get the wrapped List? Should we add a Methode for adding a interceptor.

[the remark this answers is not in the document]
alexeyzimarev: Any collection manipulation at runtime would then be required to be wrapped in locks, otherwise the client won't be thread-safe.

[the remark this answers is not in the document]
fseidl-bauradar: As fas as I see there are two ways to solve this.

RestClientOptions {
	private ImmutableList<InterceptorBase> _interceptors;

	public void AddInterceptor(InterceptorBase interceptor){
			object mutex = new object();
			lock(mutex) {
				if(_interceptors.Contains) return;
				_interceptors = _interceptors.Add(interceptor);
			}
	}
}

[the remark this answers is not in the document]
fseidl-bauradar: As I see there should be a solutiong for ensuring that the AuthenticationInterceptor is the first in the list.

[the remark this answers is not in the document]
fseidl-bauradar: Therefore I would suggest provide an option to define the Authenticator, after the client have been created.
```
class RestClientOptions {

	public void SetAuthInterceptor(IAutenticator authenticator){
		object mutex = new object;
		lock(mutex){
			Authenticator = authenticator;
		} 
	}
}
```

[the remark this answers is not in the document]
fseidl-bauradar: Do you overtake functions defined in the RestClientOptions into the RestClientReadonlyOptions?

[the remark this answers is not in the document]
alexeyzimarev: I am not sure adding mutation functions to the options class is a great idea. It's just a property bag, and the current source generator treats it like that. But I get your point. Let me think about for a bit.

src/RestSharp/RestSharp.csproj
alexeyzimarev: If you commit the change, I'll mark this as resolved. Then I will merge your PR and figure out what to do with mutability.

PR #2119, opened 2023-07-25
Improvements in processing redirects with cookie containers.

rassilon:
## Description

Improvements for #2058. This is just a starting point for these improvements, they do as you pointed out need some more tests. I have an outline for Can_Perform_RedirectingPost_With_Receive_Cookies Redirect test, I just need to make it RestSharp only as opposed to calling into the product I'm testing atm.

## Purpose
This pull request is a:

## Checklist


rassilon:
@dotnet-policy-service agree

src/RestSharp/RestClient.Async.cs
```diff
@@ -138,23 +153,65 @@ public partial class RestClient {
...
+                // fragment should inherit the fragment from the original URI.
+                string requestFragment = originalUrl.Fragment;
+                if (!string.IsNullOrEmpty(requestFragment)) {
+                    string redirectFragment = location.Fragment;
+                    if (string.IsNullOrEmpty(redirectFragment)) {
+                        location = new UriBuilder(location) { Fragment = requestFragment }.Uri;
+                    }
+                }
+
+                // Disallow automatic redirection from secure to non-secure schemes
```
alexeyzimarev, line 167: That's one thing I want to be allowed based on redirect options.

rassilon: I'm happy to help do the work if you point me in the general direction of an API surface outline. :)

alexeyzimarev: Check my comment in #2059

rassilon:
@alexeyzimarev , I'm getting a build error for HttpUtility.ParseQueryString for net472, but heck if I can understand why...

src/RestSharp/RestClient.Async.cs
rassilon: This specific decision is now driven by the new option object.

rassilon:
@alexeyzimarev , ok, I've started this in the general direction you mentioned. Still more changes to come. Let me know what you think so far. Should we have a Options.RedirectOptions property that determines the behavior of forcing the method to GET for instance?

rassilon:
@alexeyzimarev , I've updated the branch for the interceptor changes. However, should After Request be called once for the last request in the redirect chain or repeatedly for each redirect?

alexeyzimarev:
Yeah, I think so. Interceptors don't care about how the request is done. It might change in the future if we decide to add `OnRetry` and `OnRedirect`, but with the current API I think you're right, just call it when it's all done.

rassilon:
@alexeyzimarev , one thing we'll need to some of these RedirectOptions is a secure test server... Any thoughts about that fun?

rassilon:
@alexeyzimarev , I added ForwardHeaders/ForwardAuthorization/ForwardCookies tests... Should ForwardAuthorization support the Authenticator's parameter name instead of just KnownHeaders.Authorization?

rassilon:
@alexeyzimarev, I've added ForwardQuery, ForwardFragmnet, MaxRedirect, and FollowRedirectsToInsecure related tests.
This leaves ForwardBody (support/test), ForceForwardBody (test), AllowRedirectMethodStatusCodeToAlterVerb (test), and custom redirection status code additions for redirect detection (support and tests).

However, I wonder if the ForceForwardBody support is the best way to achieve that goal... Please take a look and see if you have any feedback.

rassilon:
All of the current test explorer tests pass, btw.

rassilon:
No problem. I managed to get Visual Studio .Net code coverage analysis working after I was getting nowhere with Coverlet and Fine Code Coverage. I'll be adding the rest of the test cases (indicated by coverage and otherwise), hopefully before the end of the week.

rassilon:
@alexeyzimarev, that's the last set of changes I will make before we discuss ForwardBody vs ForceForwardBody. If we should only have, say, ForceForwardBody since RedirectMethod status code (303) returns true from the helper method to determine if we should force GET instead of the current method.

I don't anticipate making any other changes before you get a chance to give me some feedback.

The code coverage looks excellent. I noticed that sometimes some of the non-"TestServer.cs" related tests fail due to trying to dispose of an already disposed server. I don't know if this is related to the xunit nuget pkg update. I also don't know if moving to IAsyncLifetime like TestServer.cs would fix them.

rassilon:
Bah, I just had this thought:
Should we ALWAYS regenerate the Cookie header after a redirect for security reasons?
And if so, should we even have ForceCookie? I'm leaning towards no.

For non-RestSharp users, if a supplied HttpClientHandler sets a CookieContainer, the HttpClient code does this as near as I can unravel from the source code. (That code makes my head hurt....) The basic idea is that the System.Net.Http inner handler of the RedirectHandler passes doAuth = false to SendAsync overloads of various sorts.

However, RestSharp doesn't supply the CookieContainer to a HttpClient HttpClientHandler, so we have to do it ourselves.

src/RestSharp/Options/RestClientOptions.cs
```diff
@@ -60,7 +66,7 @@ public class RestClientOptions {
             ? ResponseStatus.Completed
             : ResponseStatus.Error;
 
-    /// <summary>
+    /// <summary>s
```
alexeyzimarev, line 69: I guess this `s` needs to go?

alexeyzimarev:
I am not sure. I thought we should collect the response headers and pass them over to the next call. It might, indeed, create security concerns if the response contains things like authentication cookie.
I didn't look at HttpClient code to see how they handle redirects. Instead, I checked Refit and I thought we could/should do the same thing, or similar. I am wondering if you are making it too good? :)

src/RestSharp/Options/RestClientOptions.cs
rassilon: Yep. Fixed locally.

rassilon:
Hah. I think removing ForceCookie, regenerating the cookies for every redirected request (for cross-domain security reasons), and preventing the verb change without supporting ForwardBody or ForceForwardBody nonsense would be good enough. There's a reason the CookieContainer has you supply the requesting Uri for the GetCookie method after all. Hrm.. I looked at Refit briefly (https://github.com/reactiveui/refit/) but didn't see the bits that did header propagation. Did I look in the wrong place?

rassilon:
We could spin up a slack restsharp instance and have a conference call to discuss in a little more detail if you want or something similar...

alexeyzimarev:
Maybe create a Discord server?

rassilon:
> Maybe create a Discord server?

@alexeyzimarev , https://discord.com/invite/MrGRGUug (expires in 7 days)

Issue #1692, opened 2022-01-11
Client request extensions don't throw


alexeyzimarev:
Extensions like `GetAsync` or `PostAsync` should throw if the request wasn't successful.

However, it doesn't happen. The reason seems to be that the `RestResponse` instance gets the correct status (error), but doesn't get any exception set because of that. And, `ThrowIfError` only checks the exception, ignoring the status.

alexeyzimarev:
Released in 107.0.3

m1dst:
Shouldn't be closed...

m1dst:
Shouldn't be closed and should be fixed.

alexeyzimarev:
The issue is quite trivial to fix, I am wondering why no one opens a PR to get rid of this issue.

m1dst:
Don't think this ever got fixed.

alexeyzimarev:
It wasn't, but who supposes to fix it? I see you @m1dst commented quite a lot on this, have you considered spending 20 min of your time fixing it?

m1dst:
I see there is a PR that is open which is related to this.   https://github.com/restsharp/RestSharp/pull/2128

alexeyzimarev:
They weren't linked, I am checking with the PR author if it addresses this issue.

PR #2128, opened 2023-08-17
Add RestClient.DownloadStreamAsync() error handler

marcoburato-ecutek:
## Background

Our API, in case or error, returns a suitable HTTP status code and also provides more detailed information as JSON in the response body. Something like:

```json
{
    "errorCode": "CustomerNotFound",
    "errorDescription": "No customer matching the given ID was found"
}
```

For regular requests that send JSON and receive JSON, I could use `RestClient.ExecuteAsync()` to obtain a `RestResponse` and deserialise the content as either the expected object or the error object based on the HTTP status code.

When downloading files, I have found no good way of doing this. In case of error, `RestClient.DownloadStreamAsync()` would either return null or throw an exception without info about the exact server response ("Request failed with status code XXX").

### Using `DownloadStreamAsync()` + `RestRequest.OnAfterRequest`

This would work in principle. The raw response could be inspected and a suitable application-specific exception could be thrown if it's an error response.
The problem is that the consuming code would have to deal with both RestSharp responses and raw HttpClient responses, which duplicates code. RestSharp also has code to distinguish between a canceled request and a timeout, which would have to be duplicated too.

### Using `ExecuteAsync()` + `RestRequest.ResponseWriter`

This would allow to save the file to a FileStream and get a RestResponse to check the HTTP status code. The obvious problem is that, in case of error, the response error JSON would be saved to the file instead of the expected file content, which is not ideal.
Also, in case of success, RestSharp would still make a byte[] and possibly a string of the response, which is not required and undesired (especially if the file is big).

### Using `ExecuteAsync()` + `RestRequest.AdvancedResponseWriter` to save the response body

Compared to the above, this would be passed the raw HTTP response, so it has more control. But this has the same problem as RestRequest.OnAfterRequest (dealing with the raw HTTP response).

## Description

This PR adds the following overload:
`Task<Stream?> DownloadStreamAsync(RestRequest request, Action<RestResponse> errorHandler, CancellationToken cancellationToken = default)
`

If the request fails or the server returns an error HTTP status code, `errorHandler` is executed, passing the error response as a RestResponse object.
After executing the handler, the method returns null or throws an exception as usual.

Initially, I considered returning a different object which would wrap either the `Stream` or the `RestResponse`, but I didn't like the idea of returning a different type on the new overload. I also thought about using an `out RestResponse?` argument but they're not allowed on async methods.

Example usage:

```csharp
var stream = client.DownloadStreamAsync(request, (errorResponse) => {
    //Inspect response, throw application-specific exception if needed
});
```

A single error handler can be defined, to be used for many requests

```csharp
//Somewhere
var stream = client.DownloadStreamAsync(request, ThrowExceptionForErrorResponse);

//Somewhere else
var stream = client.DownloadStreamAsync(request, ThrowExceptionForErrorResponse);

void ThrowExceptionForErrorResponse(RestResponse errorResponse)
{
    //...
}
```

## Purpose
This pull request is a:

## Checklist

### DIFF
--- a/src/RestSharp/RestClient.Async.cs
+++ b/src/RestSharp/RestClient.Async.cs
@@ -40,19 +40,50 @@ public partial class RestClient {
     /// <inheritdoc />
     [PublicAPI]
     public async Task<Stream?> DownloadStreamAsync(RestRequest request, CancellationToken cancellationToken = default) {
+        return await DownloadStreamInternalAsync(request, null, cancellationToken);
+    }
+
+    /// <inheritdoc />
+    [PublicAPI]
+    public async Task<Stream?> DownloadStreamAsync(RestRequest request, Action<RestResponse> errorHandler, CancellationToken cancellationToken = default) {
+        return await DownloadStreamInternalAsync(request, errorHandler, cancellationToken);
+    }
+
+    async Task<Stream?> DownloadStreamInternalAsync(RestRequest request, Action<RestResponse>? errorHandler, CancellationToken cancellationToken = default) {
         // Make sure we only read the headers so we can stream the content body efficiently
         request.CompletionOption = HttpCompletionOption.ResponseHeadersRead;
-        var response = await ExecuteRequestAsync(request, cancellationToken).ConfigureAwait(false);
+        var internalResponse = await ExecuteRequestAsync(request, cancellationToken).ConfigureAwait(false);
 
-        var exception = response.Exception ?? response.ResponseMessage?.MaybeException();
+        if (errorHandler != null) {
+            if (internalResponse.Exception != null) {
+                var response = GetErrorResponse(request, internalResponse.Exception, internalResponse.TimeoutToken);
+
+                errorHandler(response);
+            }
+            else if (internalResponse.ResponseMessage!.IsSuccessStatusCode == false) {
+                var response = await RestResponse.FromHttpResponse(
+                    internalResponse.ResponseMessage!,
+                    request,
+                    Options.Encoding,
+                    internalResponse.CookieContainer?.GetCookies(internalResponse.Url),
+                    Options.CalculateResponseStatus,
+                    cancellationToken
+                )
+                .ConfigureAwait(false);
+
+                errorHandler(response);
+            }
+        }
+
+        var exception = internalResponse.Exception ?? internalResponse.ResponseMessage?.MaybeException();
 
         if (exception != null) {
             return Options.ThrowOnAnyError ? throw exception : null;
         }
 
-        if (response.ResponseMessage == null) return null;
+        if (internalResponse.ResponseMessage == null) return null;
 
-        return await response.ResponseMessage.ReadResponseStream(request.ResponseWriter, cancellationToken).ConfigureAwait(false);
+        return await internalResponse.ResponseMessage.ReadResponseStream(request.ResponseWriter, cancellationToken).ConfigureAwait(false);
     }
 
     static RestResponse GetErrorResponse(RestRequest request, Exception exception, CancellationToken timeoutToken) {
--- a/test/RestSharp.Tests.Integrated/DownloadFileTests.cs
+++ b/test/RestSharp.Tests.Integrated/DownloadFileTests.cs
@@ -1,5 +1,6 @@
 ﻿using System.Net;
 using System.Text;
+using RestSharp.Extensions;
 using RestSharp.Tests.Shared.Fixtures;
 
 namespace RestSharp.Tests.Integrated;
@@ -9,6 +10,7 @@ public sealed class DownloadFileTests : IDisposable {
         _server = HttpServerFixture.StartServer("Assets/Koala.jpg", FileHandler);
         var options = new RestClientOptions(_server.Url) { ThrowOnAnyError = true };
         _client = new RestClient(options);
+        _clientNoThrow = new RestClient(_server.Url);
     }
 
     public void Dispose() => _server.Dispose();
@@ -28,6 +30,7 @@ public sealed class DownloadFileTests : IDisposable {
 
     readonly HttpServerFixture _server;
     readonly RestClient        _client;
+    readonly RestClient        _clientNoThrow;
     readonly string            _path = AppDomain.CurrentDomain.BaseDirectory;
 
     [Fact]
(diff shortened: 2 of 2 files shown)


alexeyzimarev:
Thank you for the PR. Please sign the CLA, and I will merge it.

marcoburato-ecutek:
@microsoft-github-policy-service agree company="EcuTek Technologies Ltd"

marcoburato-ecutek:
@alexeyzimarev Sorry for the delay. I followed the instructions of the CLA tool to "sign" the agreement. Is that correct?

alexeyzimarev:
Doesn't seem like it. The check is still queued.

marcoburato-ecutek:
@dotnet-policy-service agree company="EcuTek Technologies Ltd"

marcoburato-ecutek:
@alexeyzimarev That worked. I looked at other PRs to see what other people did. It would be nice to mention how to accept the CLA in `CONTRIBUTING.md`.

alexeyzimarev:
Yeah, it seems like contribution guidelines need updating regardless :)

alexeyzimarev:
@marcoburato-ecutek do you think your PR fixes this one? https://github.com/restsharp/RestSharp/issues/1692

marcoburato-ecutek:
@alexeyzimarev I don't think so. My change only introduces a new method overload, it does not change previous behaviour. The assumption was that the existing behaviour was correct and intentional.

My memory is short, but according to my description above, it was possible to make `RestClient.DownloadStreamAsync()` throw on error, my problem was that it was not possible to access the information about the error (HTTP status code and message).

I reckon the other person should open a separate ticket with a test case, if they believe there's a bug.

marcoburato-ecutek:
Reopening, I thought this had been merged already.

alexeyzimarev:
There's quite a big issue here as:
1. Custom error handler is only on the client, not on the interface. Also you added `inheritdoc` but there's nothing to inherit from.
2. It only works for downloads, which creates yet another functionality gap as all other request types have no error handlers.

marcoburato-ecutek:
With regards to the interface, the documentation says
> The IRestClient interface is deprecated in v107, but brought back in v109. The new interface, however, has a much smaller API compared to previous versions. You will be using the RestClient class instance.

I interpreted that as "IRestClient is deprecated and we only partially brought it back for compatibility". So, I didn't worry about the interface. It seems I misinterpreted the documentation.

The `inheritdoc` must be my mistake.

About the fact that it only works for downloads, it's already possible to obtain the same error-handling by using `RestClient.ExecuteAsync` and checking the returned `RestResponse`. This PR was meant to address the lack of error-handling functionality when using `RestClient.DownloadStreamAsync()`.

IMHO, the best alternative solution would be to use the existing `ThrowOnAnyError` behaviour to throw an exception when the server returns an error code, which would include the HTTP status code, HTTP status message and response body (under the assumption that, for errors, it would be reasonably short). This would allow the consuming code to just use a try-catch instead of passing an error handler. However, since `RestClient` currently throws an `HttpRequestException` in case of HTTP errors (which only exposes the HTTP status code as a property), this would mean throwing a different type of exception, which would be a breaking change.

alexeyzimarev:
I have another suggestion. Why don't we extend interceptors and add `OnError` there? This way, a custom error handler can be used for any kind of request, not just downloads. It also keeps the interface and the API surface the same, so there will be no ambiguity where a `Download` with the custom error handler is only available for the implementation.

marcoburato-ecutek:
Maybe.

I think that would work _for me_, as long as the error handler is passed a `RestResponse`. But that would only allow to use a single error handler for most requests. Someone might want to have different error handling for each request.

If you add an `OnError` then you should also add an `OnSuccess` for balance. Both taking a `RestResponse` instead of an `HttpResponseMessage`. `InterceptBeforeDeserialize` would not be the same as it would not be called for a download, I imagine.

What about having a method overload like `Task<Stream?> DownloadStreamAsync(RestRequest request, out RestResponse? errorResponse)`, where on success errorResponse is null and a stream is returned, while on error errorResponse is non-null and null is returned? That would make error handling code closer to `ExecuteAsync()`: no lambda/interceptor, the code gets a RestResponse and inspects it.

alexeyzimarev:
The problem with `out` would be that, first, it still requires signature change, second, it forces people to add an out var even if they don't need it.

`InterceptBeforeDeserialize` indeed is only called in generic extensions, when deserialisation indeed takes place.

I would remove it, rename `InterceptBeforeRequest` and `InterceptAfterRequest` to `InterceptBeforeHttpRequest` and `InterceptAfterHttpRequest`. Then, add `InterseptAfterSuccessfulRequest(RestResponse reponse)` and `InterseptAfterFailedRequest(RestResponse reponse)`. The last one would give you the error response.

alexeyzimarev:
Adding stuff to the interface breaks all custom interface implementations as people would need to add this new method there even if they don't use it.

alexeyzimarev:
Another thing is that the error response example you provided when you opened the PR is quite common. It's not only limited to downloads. Often, normal calls return similar results or, for example, problem details. Ultimately, the solution could be that the response deserialisation would support two types. 

But, indeed, the download case is different because you expect a stream back, but there's nothing there because where's an error. Still, the implementation in this PR would still produce an exception unless you disable `ThrowOnAnyError`, and you have a completely separate flow for error handlings, which is still quite inconvenient (interceptors won't help there though).


I guess the best solution for downloads would be to change the signature of `DownloadStreamAsync` to return a tuple of `Stream` and `TError?`. RestSharp can then auth-deserialise the response to `TError`, which would be convenient to inspect in the same execution path instead of separate function/interceptor. The interface will be broken, but it will be easy to fix downstream. Also, it will be possible to retrofit the old API there by an extension with the current signature, which will deserialise errors to `JObject` a not return it. So only the code that implements the interface will be broken, but, again, it's an easy fix.

marcoburato-ecutek:
Regarding the interface, my understanding is that it's possible to add [default implementations](https://devblogs.microsoft.com/dotnet/default-implementations-in-interfaces/) to avoid breaking existing consumers. But, I've never used the feature myself and I don't know if it's compatible with old runtimes.

Following your idea, I reckon the signature on the interface should be a tuple of `Stream?, RestResponse?`. Automatically deserializing an error object could then be done in the same way ExecuteAsync does it (extension methods, if I understand correctly). That way one could deal with the low-level RestResponse if needed.
For instance, right now I'm dealing with another API that gives a plain `503 Service Unavailable` error response that can include a `Retry-After` header, which is of interest. I don't need to use RestSharp in this case, just providing it as an example.

DontEatRice:
Any update on this? 

> I guess the best solution for downloads would be to change the signature of DownloadStreamAsync to return a tuple of Stream and TError?

I don't like this approach, since I will not be able to read the status code, response headers that often include VERY useful information like file content-type or file name. In Blazor we are now using this new thing called interceptors, but to be honest, right now it is more usefull to use raw HttpClient for file downloading scenario, because at least we are getting a HttpResponse easily

alexeyzimarev:
@marcoburato-ecutek thank you for this PR and for the patient design discussion. Reading through the thread again, the core problem you raised — that downloads have no way to inspect the response on failure — is genuinely real and worth solving, but I've come around to the view that the callback-overload shape isn't quite the right primitive. Specifically:

- It only addresses `DownloadStreamAsync`, leaving `DownloadDataAsync` and any future streaming method with the same gap.
- It's on `RestClient` only, not `IRestClient`, so anyone consuming via the interface (the recommended pattern) can't use it.
- The `Action<RestResponse>` callback forces every caller to introduce a captured local and check it after the call — the tests in this PR show that awkwardness.

I've opened **#2386** with a proper design proposal: a new `IDownloadClient` companion interface plus a `DownloadResult` record that carries both the `Stream` and the full `RestResponse`. The existing `DownloadStreamAsync` gets reimplemented in terms of the new path with no behaviour change, so nothing breaks. That solves your original use case (inspecting the error body), @DontEatRice's case (reading content-type/filename headers), and the `Retry-After` scenario you mentioned later in this thread, all in a single return value.

Closing this PR in favour of #2386. The design discussion here was load-bearing for the new proposal — thanks again for pushing on the problem and for the test cases, which I'll likely lift wholesale when implementing.

PR #2141, opened 2023-09-14
Wrapping up interceptors, fixing typos, removing Moq

alexeyzimarev:
- Make interceptors collection immutable in read-only options
- Adding interceptors to request level, making old callbacks obsolete
- Refactoring tests to not use Moq
- Moving `OnBeforeDerialization` to the right place
- Aligning interceptors:
  - Before request
  - Before HTTP request
  - After HTTP request
  - After request
  - Before deserialization

### DIFF
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -20,8 +20,17 @@ jobs:
           dotnet-version: '7.0'
       -  
         name: Run tests
-        run: dotnet test -c Release
-        
+        run: dotnet test
+      -
+        name: Publish test results
+        uses: EnricoMi/publish-unit-test-result-action@v2
+        if: always()
+        with:
+          files: |
+            test-results/**/*.xml
+            test-results/**/*.trx
+            test-results/**/*.json
+
   docs:
     runs-on: ubuntu-latest
 
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -1,0 +1,37 @@
+<Project>
+  <PropertyGroup>
+    <ManagePackageVersionsCentrally>true</ManagePackageVersionsCentrally>
+  </PropertyGroup>
+  <PropertyGroup Label="Package versions for .NET 6" Condition="$(TargetFramework) == 'net6.0'">
+    <MicrosoftTestHostVer>[6.0.22,7)</MicrosoftTestHostVer>
+  </PropertyGroup>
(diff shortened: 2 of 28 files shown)


alexeyzimarev:
@fseidl-bauradar you might want to take a look

alexeyzimarev:
I am not sure if throwing in interceptors should crash the request, like it's expected in tests. The current OnXXX hooks can produce exceptions, and if that results in client throwing depends on throw options. I want to replace the current hooks with interceptors, so I think they should express the same behaviour.

fseidl-bauradar:
> I am not sure if throwing in interceptors should crash the request, like it's expected in tests. The current OnXXX hooks can produce exceptions, and if that results in client throwing depends on throw options. I want to replace the current hooks with interceptors, so I think they should express the same behaviour.

In my project, I am currently using it to convert a ErrorDTO from the DTO into a Exception.
Because my Application is currently in a HybridMode State, between using REST and direct DatabaseAccess.
But as I think the behaviour that a Exception in the Interceptors cancels the Request, sounds logical.
In Special when you are adding CancelationTokens to the interceptors

alexeyzimarev:
Yeah, I didn't think that interceptors _can't_ cancel the request, but the way it work now with hooks is that the `ExecuteAsync` doesn't throw unless `ThrowOnAnyError` is set to `true` and you can an error response back where the `RestResponse.ErrorException` is set to the produced exception.

fseidl-bauradar:
> Yeah, I didn't think that interceptors _can't_ cancel the request, but the way it work now with hooks is that the `ExecuteAsync` doesn't throw unless `ThrowOnAnyError` is set to `true` and you can an error response back where the `RestResponse.ErrorException` is set to the produced exception.

Sorry, but your answer is a little bit confusing, 
As I understand it you want to say that interceptors can't cancel the request, but the way it work now with hooks is that the ExecuteAsync doesn't throw unless `ThrowOnAnyError` is set to `true` and a developer get a resppnse back where the `RestResponse.ErrorException` is set to the produces exception.

Am I understanding you right?
One possible usage of an interceptor, would be a validation of the data before sending it to the Server, avoiding useless roundtrips.
Is this possible to handle by your implementation?

fseidl-bauradar:
Have you found the reason for the build on linux and Claudflare Pages?
Would be nice if I could use the official package again and don't use my self hosted version for ever.

PR #2146, opened 2023-10-13
Use case insensitive comparer

softworkz:
..instead of ToLower()


softworkz:
@dotnet-policy-service agree

PR #2157, opened 2023-11-03
Set UserAgent as a default header parameter

PetesBreenCoding:
## Description
When a UserAgent is set in the RestClientOptions passed to the RestClient constructor, the UserAgent is now set as a HTTP header in the DefaultParameters property rather than modifying the HttpClient instance. This is so that the HttpClient instance can be re-used for APIs etc that require different UserAgent headers.

The ConfigureHttpClient method still sets the max timeout as that's set on the client instance itself, not a HTTP header.

I did not modify the "ExpectContinue" behaviour as it seems to have complex logic around parsing values rather than just directly setting them. If this is something you'd like to be in the DefaultParameters property too I can consider that.

## Purpose
This pull request is a:

## Checklist

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -594,7 +594,7 @@ One way of doing it is to use `RestClient` constructors that accept an instance 
 
 - `BaseAddress` will be used to set the base address of the `HttpClient` instance if base address is not set there already.
 - `MaxTimeout`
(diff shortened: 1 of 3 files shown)


PetesBreenCoding:
> `@dotnet-policy-service agree [company="{your company}"]`

@dotnet-policy-service agree

Issue #2169, opened 2024-02-02
Update System.Text.Json to 8.0.0+


larschristensen20:
**DO NOT USE ISSUES FOR QUESTIONS**

**Is your feature request related to a problem? Please describe.**
Using RestSharp in a .Net Framework 4.8 windows client, to query our ASP.net API hosted in Azure. Ran into a problem with a large (>125MB) query. [https://github.com/dotnet/runtime/issues/39953#issuecomment-1772516750](Described here). 
When I install RestSharp 110+ it installs System.Text.Json 7.0.2 as well. However, updating System.Text.Json to 8.0.1 for example, does not seem to remove the limitation, since I assume RestSharp internally refers to a different version of System.Text.Json? Correct me if I'm wrong.

**Describe the solution you'd like**
It seems a fix to the problem of serializing more than 125MB of base 64 data is to use the newest version of System.Text.Json. I would like to either A) be able to use System.Text.Json 8.0.0+ with RestSharp 110 by installing it as a separate nuget package, or B) have RestSharp 110+ updated to use it internally.

**Describe alternatives you've considered**
I've considered using Newtonsoft as serializer, but I'm bummed over losing the performance of System.Text.Json serializers. Also considered making a different set of endpoints that specifically uses Newtonsoft, but I'd like to avoid it since I imagine it can introduce some problems.

alexeyzimarev:
I think it's better to solve it with #2177

PR #2173, opened 2024-03-18
.Net 8.0 Compliance

TTWill:
## Description

To have Net 8.0 compliance

## Purpose
This pull request is addressing #2169 

- [ ] Bugfix (non-breaking change "should" fix compliance with .Net 8.0)
- [ ] Conflicting issue with test **Can_Add_Object_Static_with_custom_property_name**: "Should be Hello world,Guid[] Array" , but "Hello world,Guid[]-matrix" is returned, likely due to updated language behavior.
- [ ] Conflicting issue with test **Task_Handles_Non_Existent_Domain** is depending on culture-biased result. Recommended practice is testing deterministic data and to avoid testing message strings.

### DIFF
--- a/src/RestSharp.Serializers.CsvHelper/RestSharp.Serializers.CsvHelper.csproj
+++ b/src/RestSharp.Serializers.CsvHelper/RestSharp.Serializers.CsvHelper.csproj
@@ -1,4 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
+    <PropertyGroup>
+      <TargetFrameworks>netstandard2.0;net471;net6.0;net7.0;net8.0</TargetFrameworks>
+    </PropertyGroup>
     <ItemGroup>
         <PackageReference Include="CsvHelper" Version="30.0.1" />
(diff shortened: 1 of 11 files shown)


TTWill:
@dotnet-policy-service agree company="Microsoft"

TTWill:
In one of the commits to follow, I changed one of the tests to take into account, that `Array.Empty<Guid>()`  gave a different output. It turns out this was due to **System.Text.Json** being v7.0.3 for net4.7.2 - this appears to be related to the issue #2169 - so, changing all frameworks to use v8.0.3 appear to be eliminating this issue, thus all tests could be completed accordingly.

Note: Another test was also changed, since it renders a **false negative** due to me using culture specific system (da-DK). Details is enclosed within the commit.

Finally, when using this version, I advice to **CLEAN** your solution or the #2169 issue may remain.

alexeyzimarev:
I am not sure what is the difference between this and #2174 as code changes are mostly the same, except of the test change. That's why I suggest opening a separate PR to fix that test. Thank you.

PR #2174, opened 2024-03-19
.Net 8.0 Compliance Followup

### LINKED ISSUE #2169: Update System.Text.Json to 8.0.0+

TTWill:
## Purpose
To add full .Net 8.0 compliance and and fix #2169 

## Description
In one of the commits here, I changed one of the tests to take into account, that `Array.Empty<Guid>()`  gave a different output. It turns out this was due to **System.Text.Json** being v7.0.3 for net4.7.2 - this appears to be related to the issue #2169 - so, changing all frameworks to use v8.0.3 appear to be eliminating this issue, thus all tests could be completed accordingly.

_Note_: Another test was also changed, since it renders a **false negative** due to me using culture specific system (da-DK). Details is enclosed within the commit.

Finally, when using this version, I advice to **CLEAN** your solution or the #2169 issue may remain.

### DIFF
--- a/src/RestSharp.Serializers.CsvHelper/RestSharp.Serializers.CsvHelper.csproj
+++ b/src/RestSharp.Serializers.CsvHelper/RestSharp.Serializers.CsvHelper.csproj
@@ -1,4 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
+    <PropertyGroup>
+      <TargetFrameworks>netstandard2.0;net471;net6.0;net7.0;net8.0</TargetFrameworks>
+    </PropertyGroup>
     <ItemGroup>
         <PackageReference Include="CsvHelper" Version="30.0.1" />
     </ItemGroup>
--- a/src/RestSharp.Serializers.NewtonsoftJson/RestSharp.Serializers.NewtonsoftJson.csproj
+++ b/src/RestSharp.Serializers.NewtonsoftJson/RestSharp.Serializers.NewtonsoftJson.csproj
@@ -1,4 +1,7 @@
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
+    <PropertyGroup>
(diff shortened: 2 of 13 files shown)


TTWill:
@dotnet-policy-service agree company="Nordicscreen ApS"

alexeyzimarev:
Target frameworks, including `net8.0`, are in the props file. There's no need to add those tags and groups to all the projects.

alexeyzimarev:
I prefer target frameworks to stay where they are.
Also, this line:
```
<HintPath>..\..\..\..\Program Files (x86)\Reference Assemblies\Microsoft\Framework\.NETFramework\v4.7.2\System.Web.dll</HintPath>
```
Makes a bold assumption about where the referenced assembly is. It's not going to work.

I updated STJ to 8.0.3 in another PR, and removed it as a dependency for newer .NET versions.

For the test you changed, which looks at the exception message and fails on non-English locales, I would suggest making a separate PR for it.

PR #2204, opened 2024-05-24
Rename HttpBasicAuthenticator to HttpBasicAuth in usage docs

ChayoteJarocho:
## Description

Fix a small typo, as the former type does not exist but the latter does.

## Purpose
This pull request is a:

## Checklist


ChayoteJarocho:
@dotnet-policy-service agree

alexeyzimarev:
Sorry, the rename was not intentional and will be reverted, so there's no need to change the docs.

PR #2265, opened 2024-09-18
Allow leaving encoded slash in UrlSegmentParameter value

cg-finbourne:
## Description

We need to use path parameters which include the sequence of characters '%2F' or '%2f' in them. The constructor of UrlSegmentParameter replaces these sequences in the value with a '/' character. When using version 110.2.0 of RestSharp we would get around this by setting the Value property with the original string after the creation of the UrlSegmentParameter, but at 112.0.0 this is no longer possible as the Value property is now readonly after creation.

This pull request adds an optional parameter to the UrlSegmentParameter constructor to give control over whether the value is changed to replace '%2F' or not.

(we cannot create a new class with the desired behaviour since the UrlSegmentParameter type is used in this line https://github.com/restsharp/RestSharp/blob/dev/src/RestSharp/Request/UriExtensions.cs#L57)

## Purpose
This pull request is a:

## Checklist

### DIFF
--- a/src/RestSharp/Parameters/UrlSegmentParameter.cs
+++ b/src/RestSharp/Parameters/UrlSegmentParameter.cs
@@ -26,10 +26,11 @@ public partial record UrlSegmentParameter : NamedParameter {
     /// <param name="name">Parameter name</param>
     /// <param name="value">Parameter value</param>
     /// <param name="encode">Optional: encode the value, default is true</param>
-    public UrlSegmentParameter(string name, string value, bool encode = true)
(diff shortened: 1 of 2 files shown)


cg-finbourne:
@dotnet-policy-service agree [company="finbourne"]

cg-finbourne:
@dotnet-policy-service agree company="finbourne"

cg-finbourne:
I have created this issue for it https://github.com/restsharp/RestSharp/issues/2267

kenliao43088 (approved):

